PR #4, opened 2020-10-27
Build stripped binaries to reduce sizes

n3wscott:
ref: https://github.com/knative/test-infra/pull/2537


mattmoor (commented):
/lgtm

cc @tcnghia

PR #5, opened 2020-10-28
Move tests to the hack repo

chizhg:
Also move the test files to the hack repo.

/cc @n3wscott

### DIFF
--- a/hack.go
+++ b/hack.go
@@ -1,3 +1,5 @@
+// +build hack
+
 /*
 Copyright 2020 The Knative Authors
 
--- a/hack.go
+++ b/hack.go
@@ -1,5 +1,3 @@
-// +build hack
-
 /*
 Copyright 2020 The Knative Authors
 
--- a/test/e2e/simple_e2e_test.go
+++ b/test/e2e/simple_e2e_test.go
@@ -1,0 +1,29 @@
+// +build e2e
+
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"testing"
+)
+
+// This is actually a unit test, but it does exercise the necessary helper functions.
+
+func TestE2ESucceeds(t *testing.T) {
+	// Always succeed.
+}
--- a/test/library_test.go
+++ b/test/library_test.go
@@ -1,0 +1,59 @@
+// +build library
+
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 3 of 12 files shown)


test/unit/presubmit-tests.sh
```diff
@@ -0,0 +1,281 @@
...
+    (( ! POST_UNIT_TESTS )) || test_failed "Post unit tests did run"
+    (( ! PRE_INTEGRATION_TESTS )) || test_failed "Pre integration tests did run"
+    (( ! RAN_INTEGRATION_TESTS )) || test_failed "Custom integration tests did run"
+    (( ! POST_INTEGRATION_TESTS )) || test_failed "Post integration tests did run"
+    echo "Test failed fast"
+  }
+}
+
+function test_exempt_md() {
+  source $(dirname $0)/../../scripts/presubmit-tests.sh
```
n3wscott, line 190: This is no longer in the scripts dir.

Search the rest of the code for scripts?

test/unit/presubmit-tests.sh
```diff
@@ -0,0 +1,281 @@
...
+PROW_JOB_ID=123
+JOB_TYPE="presubmit"
+ARTIFACTS=/tmp
+
+source $(dirname $0)/test-helper.sh
+
+set -e
+
+function init_test_env() {
+  source $(dirname $0)/../../scripts/presubmit-tests.sh
```
n3wscott, line 29: And here

n3wscott (commented):
Some missed paths on scripts

test/unit/presubmit-tests.sh
chizhg: Done.

test/unit/presubmit-tests.sh
chizhg: Done.

chizhg:
Please note I had to remove `//+build hack` to make `go test ./...` not panic.

hack.go
```diff
@@ -1,5 +1,3 @@
-// +build hack
```
n3wscott, line 1: We really do want to have this tag here.

test/e2e/simple_e2e_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+package e2e
+
+import (
+	"testing"
+)
+
+// This is actually a unit test, but it does exercise the necessary helper functions.
+
+func TestE2ESucceeds(t *testing.T) {
```
n3wscott, line 27: What is the point of these? We should drop them from this repo. They should not be in the hack repo.

test/library_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"fmt"
+	"os"
+	"syscall"
+	"testing"
+	"time"
+)
+
+// signal sends a UNIX signal to the test process.
+func signal(s os.Signal) {
```
n3wscott, line 30: What is the point of these? We should drop them from this repo. They should not be in the hack repo.

chizhg: This is for unit testing the `library.sh` script.

test/e2e/simple_e2e_test.go
chizhg: This is for the smoke e2e tests to make sure the base e2e test flow is not broken.

hack.go
chizhg: What's the reason behind it? We had https://github.com/knative/test-infra/blob/master/scripts/dummy.go which doesn't have build tags.

n3wscott: It was a mistake to not have build tags on the original file. We are using this file for go mod control, but no code should be here.

chizhg: Got it, I have added it back.

n3wscott (commented):
/lgtm
/approve

PR #7, opened 2020-10-29
hack should have a hack folder... and changed copyright years

maximilien:
1. `hack/hack` suggestions and reorg since makes it easy to reuse when setting up sandbox projects
2. updated copyrights to 2020


coryrc:
These files are old so their copyrights should reflect their original creation date

coryrc:
also these scripts are intended for others to use so don't belong in under a hack directory

maximilien:
Interesting... I looked at it as here’s a template to make your project’s `hack` creation easier and streamlined with everyone else.

coryrc:
Hmm, we should've thought about that during name creation!
It's sole purpose is to hold the test-infra/scripts directory somewhere else

maximilien:
NP @coryrc. I’ll close this then. Cheers

PR #8, opened 2020-10-30
make sure to use a release family and not just master.

n3wscott:
A release lead will need to bump the default until knobots can set it.


n3wscott:
/hold

This... might not be what I want to do...

PR #10, opened 2020-11-02
do not use the sum.golang.org service, too many 410s

n3wscott:
We are still getting a lot of 410's when doing update deps, we are using direct go mod mode, but there is a second checksum service that is also in-play. Removing that service.

Ref: https://twitter.com/mattomata/status/1323296883838627840

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -512,6 +512,7 @@ function go_update_deps() {
 
   export GO111MODULE=on
   export GOFLAGS=""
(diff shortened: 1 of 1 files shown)


mattmoor (commented):
/lgtm

dprotaso:
I'm worried people will forget this is turned off and not realize we're exposing ourselves

Can we add retries for CI?

PR #16, opened 2020-11-06
Do not output json files since the Golang issue has been fixed

chizhg:
This was added in https://github.com/knative/test-infra/pull/2348, and we don't need it anymore since the Golang issue has been fixed.


n3wscott (commented):
/lgtm
/approve

PR #17, opened 2020-11-06
the branch name is no longer a valid default for release

n3wscott:


library.sh
```diff
@@ -517,7 +517,7 @@ function go_update_deps() {
   echo "=== Update Deps for Golang"
 
   local UPGRADE=0
-  local VERSION="master"
+  local VERSION="v9000.1" # release v9000 is so far in the future, it will always pick the default branch.
```
grantr, line 520: itsover1000

PR #19, opened 2020-11-12
chore: filter out bogus warnings about symlinks when doing a go mod tidy

rhuss:
This PR shuts up spurious warnings in the client build, but still
let through other warnings.

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -554,8 +554,11 @@ function go_update_deps() {
   echo "--- Go mod tidy and vendor"
 
   # Prune modules.
+  local orig_pipefail_opt=$(shopt -p -o pipefail)
+  set -o pipefail
   go mod tidy 2>&1 | grep -v "ignoring symlink" || true
   go mod vendor 2>&1 |  grep -v "ignoring symlink" || true
+  eval "$orig_pipefail_opt"
 
   echo "--- Removing unwanted vendor files"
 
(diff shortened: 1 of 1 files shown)


library.sh
```diff
@@ -554,8 +554,8 @@ function go_update_deps() {
   echo "--- Go mod tidy and vendor"
 
   # Prune modules.
-  go mod tidy
-  go mod vendor
+  go mod tidy 2>&1 | grep -v "ignoring symlink" || true
+  go mod vendor 2>&1 |  grep -v "ignoring symlink" || true
```
markusthoemmes, line 558: I did try a similar thing because they annoyed me, but this will always return successfully, right? Even if there's issues with tidy/vendor?

Maybe we should capture the output, filter it, print it but exit with the original return code.

rhuss: I like to have live output as kind of a progress meter. Maybe let's `set -o pipefail` before the call, just to break the script in case of an error (we could reset it afterwards, but `set -o pipefail` is always a good thing). Or maybe it's already set ?

BTW, that annoying message is a go issue described in https://github.com/golang/go/issues/35941

rhuss: @markusthoemmes I added a temporary `set -o pipefail` which should cause the pipe to fail when go mod tidy/vendor fails.

n3wscott (commented):
/lgtm
/approve

PR #22, opened 2020-11-16
Don't overwrite KO_DOCKER_REPO if already set

### LINKED ISSUE #21: Don't overwrite KO_DOCKER_REPO if set already
We shouldn't overwrite `KO_DOCKER_REPO` env variable if it is already set in current shell. Users might want to use their own registry for whatever reason.

Currently, `KO_DOCKER_REPO` is overwritten if we are using GKE based cluster and running:

```
test/e2e-tests.sh --run-tests
```

cardil:
Fixes #21 

### Changes
- Don't overwrite `KO_DOCKER_REPO` if already set in current shell, and using GKE.

### DIFF
--- a/e2e-tests.sh
+++ b/e2e-tests.sh
@@ -69,8 +69,12 @@ function setup_test_cluster() {
   # Acquire cluster admin role for the current user.
   acquire_cluster_admin_role "${k8s_cluster}"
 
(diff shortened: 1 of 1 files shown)


n3wscott (commented):
/lgtm
/approve

PR #23, opened 2020-11-18
Add latest_version function

pierDipi:
Eventing and Serving define this function in their test scripts.

It could be useful for other projects as well.

References:
- https://github.com/knative/eventing/blob/12b783a7f2e7fbc54d20af56a53af80476e68df2/test/e2e-common.sh#L68-L74
- https://github.com/knative/serving/blob/631927b80dbd80c2ca6dc45b2638781ed805f919/test/e2e-upgrade-tests.sh#L34

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -741,6 +741,14 @@ function shellcheck_new_files() {
   fi
 }
 
+function latest_version() {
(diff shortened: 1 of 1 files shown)


n3wscott (commented):
/lgtm
/approve

PR #25, opened 2020-11-19
Install imc and mt channel broker manifests

navidshaikh:
with knative nightly and released version install utils

follow up PR to #24 

The eventing.yaml is [brought back already](https://github.com/knative/eventing/pull/4546), so we can consider reverting #24 or keep it this way. wdyt? /cc @n3wscott @chizhg

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -412,6 +412,8 @@ function start_latest_knative_serving() {
 # Install Knative Eventing in the current cluster.
 # Parameters: $1 - Knative Eventing crds manifest.
 #             $2 - Knative Eventing core manifest.
+#             $3 - Knative Eventing in memory channel manifest.
+#             $4 - Knative Eventing mt channel broker manifest.
 function start_knative_eventing() {
   header "Starting Knative Eventing"
   subheader "Installing Knative Eventing"
@@ -419,6 +421,10 @@ function start_knative_eventing() {
   kubectl apply -f "$1"
   echo "Installing Eventing core components from $2"
   kubectl apply -f "$2"
+  echo "Installing Eventing in memory channel from $3"
+  kubectl apply -f "$3"
+  echo "Installing Eventing mt channel broker from $4"
+  kubectl apply -f "$4"
   wait_until_pods_running knative-eventing || return 1
 }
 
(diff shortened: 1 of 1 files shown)


library.sh
```diff
@@ -412,26 +412,34 @@ function start_latest_knative_serving() {
...
   header "Starting Knative Eventing"
   subheader "Installing Knative Eventing"
   echo "Installing Eventing CRDs from $1"
   kubectl apply -f "$1"
   echo "Installing Eventing core components from $2"
   kubectl apply -f "$2"
+  echo "Installing Eventing in memory channel from $3"
+  kubectl apply -f "$3"
+  echo "Installing Eventing mt channel broker from $4"
+  kubectl apply -f "$4"
```
matzew, line 427: Are we sure we want them here? 

Is there a different user of the `start_knative_eventing` function that is expecting IMC and channel broker to be NOT there?  can we leverage those installs in separate functions?

library.sh
```diff
@@ -412,26 +412,34 @@ function start_latest_knative_serving() {
...
 }
 
 # Install the stable release Knative/eventing in the current cluster.
 # Parameters: $1 - Knative Eventing version number, e.g. 0.6.0.
 function start_release_knative_eventing() {
   start_knative_eventing "https://storage.googleapis.com/knative-releases/eventing/previous/v$1/eventing-crds.yaml" \
-    "https://storage.googleapis.com/knative-releases/eventing/previous/v$1/eventing-core.yaml"
+    "https://storage.googleapis.com/knative-releases/eventing/previous/v$1/eventing-core.yaml" \
+    "https://storage.googleapis.com/knative-releases/eventing/previous/v$1/in-memory-channel.yaml" \
+    "https://storage.googleapis.com/knative-releases/eventing/previous/v$1/mt-channel-broker.yaml"
```
matzew, line 437: I wonder why there were not there - I guess I lack context here

navidshaikh: @matzew : I raised this https://github.com/knative/hack/pull/24 yesterday with only -core and -crds and missed the imc and mt-broker, fixing that with this follow up.

navidshaikh: but since `eventing.yaml` sums that together and is back now (https://github.com/knative/eventing/pull/4546) , I wonder if we should (revert) use that instead of these 4 yamls.

navidshaikh:
alternate to #26 

/hold

navidshaikh:
/close

I think installing eventing using single yaml (eventing.yaml) is better than via 4 yamls. 
#26 supersedes this.

PR #27, opened 2020-11-19
Add microbenchmarks scripts

slinkydeveloper:
These are taken from https://github.com/knative/eventing/tree/master/test.
The goal is to add later a github action to run them, so we can easily support running microbenchmarks in all the different knative projects.

Sorry in advance for my poor shell scripting abilities.

Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>


n3wscott:
I think the failure is your scripts should define a set of bash functions that integrators use by importing them, not direct run scripts. 

See https://github.com/knative/hack/blob/master/e2e-tests.sh as an example. Note none of the other hack scripts are something you run, only something you import.

slinkydeveloper:
@n3wscott How should I use them in my GH actions? Should I source them and run the single functions?

slinkydeveloper:
That should do it @n3wscott

n3wscott (commented):
/lgtm
/approve

PR #30, opened 2020-11-24
Shell out package

### LINKED ISSUE #28: API to call out shell scripts & functions from within Golang
## Summary

Write an API to enable calling out shell scripts or functions from within hack or test Golang code.

## Why?

We would like to write more code in Golang, that now written in shell scripts. That involves environment orchestration, test frameworks, build tools and hacks. To make that happen we need a bridge solution which let us call out working shell scripts, and not force us to rewrite everything at once.

Such proposals were previously discussed on Productivity WG https://github.com/knative/test-infra/issues/2418

There are, also other use cases that can benefit from this. Like, new upgrade test framework in which we would like to install either latest stable release or current git HEAD version. Those are now written as Bash shell functions. Such shell call outs can be seen implemented in knative/eventing#4519 [`test/upgrade/installation/git_head.go L25-L37`](https://github.com/knative/eventing/pull/4519/files#diff-af7a4315e01a561183b5ea793125285f664c46525bce2714cebffa48cf595231R25-R37) Implementing that in knative/pkg is better idea then doing that for eventing, serving, contrib... etc.

## How?

We should write an API that will enable us to simple call out shell scripts or just ad-hoc shell functions. It should have a logic to easily let users call out shell files relative to a given Golang source file, so they can invoke project or vendored scripts.

Invoking that script should result in STDOUT and STDERR streamed to Golang process `os.Stdout` and `os.Stderr`. 

It would be nice to prefix those streams with some labels to make that output more readable.

cardil:
Fixes #28 

This shell out package should be treated as a bridge solution that will help Knative move out of using too much shell scripts. It will enable us using new Golang orchestrated tests, and build processes, and have the ability to shell out to scripts if needed.

### DIFF
--- a/shell/assertions_test.go
+++ b/shell/assertions_test.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package shell_test
+
+import (
+	"strings"
+	"testing"
+)
+
+type assertions struct {
+	t *testing.T
+}
+
+func (a assertions) NoError(err error) {
+	if err != nil {
+		a.t.Error(err)
+	}
+}
+
+func (a assertions) Contains(haystack, needle string) {
+	if !strings.Contains(haystack, needle) {
+		a.t.Errorf("wanted to \ncontain: %#v\n     in: %#v",
+			needle, haystack)
+	}
+}
+
+func (a assertions) Equal(want, got string) {
+	if got != want {
+		a.t.Errorf("want: %#v\n got:%#v", want, got)
+	}
+}
--- a/shell/executor.go
+++ b/shell/executor.go
@@ -1,0 +1,189 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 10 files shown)


cardil:
You can see a usage of this code in https://github.com/knative/eventing/pull/4519 - here exactly [knative/eventing#4519 test/upgrade/installation/shell.go@L24-L38](https://github.com/knative/eventing/pull/4519/files#diff-1e6fb7b3d1e6b37573b1f6465bbe09dd626e41f523d7a05086456bc70f08af15R24-R38).

Here is an upgrade test prow job, in which those shell out can be seen: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_eventing/4519/pull-knative-eventing-upgrade-tests/1331366542394265600

shell/assertions_test.go
```diff
@@ -0,0 +1,45 @@
...
+		a.t.Errorf("wanted to \ncontain: %#v\n     in: %#v",
+			needle, haystack)
+	}
+}
+
+func (a assertions) Equal(want, got string) {
+	if got != want {
+		a.t.Errorf("want: %#v\n got:%#v", want, got)
+	}
+}
```
slinkydeveloper, line 45: Use testify for that?

cardil: I wanted to avoid adding dependencies, as that opens a pandora box, of whether to vendor them or not.

slinkydeveloper: I guess this pandora box has to be opened at some point right?

cardil: Not necessarily. In the ideal world I would see **ALL** contents of this repo, gradually being deprecated, and replaced by dedicated tools, and Golang code.

slinkydeveloper: exactly, to develop this golang code (eg embed other tools) you'll probably need managing dependencies anyway right?

cardil: Yeah, But maybe in other repos then knative/hack?!? IDK

n3wscott (approved):
/lgtm
/approve

/hold 
Can you as a follow up or another commit to mark every exported method as `// Deprecated: do not use this` to avoid profiliation? I am hoping downstreams seeing the deprecated method in use will encourage other solutions.

Unhold if you want to do it as a followup. Thanks!!

cardil:
Thanks, @n3wscott! 

I'll do that as a follow up. I have an idea, of how to do it even more strict.

/unhold

PR #33, opened 2020-11-30
Create a go path that points back to the local directory.

### LINKED ISSUE #15: Let us code-gen off the GOPATH
See: https://github.com/knative-sandbox/knobots/blob/master/.github/workflows/auto-updates.yaml#L165-L170

We should add this to the update codegen scripts.

n3wscott:
Fixes #15 

This change will provide a hook to integrate with out-of-GOPATH projects to use k8s codegen and the changes are redirected back to the module.

Example integration in https://github.com/knative-sandbox/discovery/pull/125

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -578,15 +578,14 @@ function go_update_deps() {
 # Return a GOPATH to a temp directory. Works around the out-of-GOPATH issues
 # for k8s client gen mixed with go mod.
 # indended to be used like:
-#   export GOPATH=$(go_mod_path)
-function go_mod_path() {
+#   export GOPATH=$(go_mod_gopath_hack)
+function go_mod_gopath_hack() {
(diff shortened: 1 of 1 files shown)


library.sh
```diff
@@ -575,6 +575,20 @@ function go_update_deps() {
   remove_broken_symlinks ./vendor
 }
 
+# Return a GOPATH to a temp directory. Works around the out-of-GOPATH issues
+# for k8s client gen mixed with go mod.
+# indended to be used like:
+#   export GOPATH=$(go_mod_path)
+function go_mod_path() {
+	export MODULE_NAME=$(go mod graph | cut -d' ' -f 1 | grep -v '@' | head -1)
```
n3wscott, line 583: use spaces.

mattmoor (commented):
/lgtm

PR #34, opened 2020-11-30
adding github aware tags to new headers

n3wscott:
Starts to implement https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions


mattmoor (commented):
/lgtm

PR #37, opened 2020-12-14
Use testname format for gotestsum command

chizhg:
This will significantly reduce the size of the JUnit XML files produced by `gotestsum`. And note it will still include all the log messages for the failed test cases, so it does not impact the debuggability when the test flows fail.


fejta (approved):
/hold

Thanks!

PR #43, opened 2021-01-13
Use git tag directly in master to get latest version

### LINKED ISSUE #45: latest_version might not always work in main/master branch
The current `latest_version` uses `git describe` to find the latest major and minor version. This is necessary to work with PRs to older release branches. However, when there is a release error and new commits are needed the auto release Prow job will add the release tag to the HEAD of the release branch instead of the root of the branch. As a result, `git describe` in master/main cannot find the release tag using `git describe`. For example, event 0.20 is released in eventing-kafka, `git describe` will show `0.19`, which will break the upgrade test or any tests depending on this function.

zhongduo:
Fix #45 

# Changes

`git describe` does not always work properly in all the repos,
especially when hot fix is required for the release process. After
this commit, when the PR is for master it will simply use git tag
without specifying the major version to get the latest tag instead.

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -794,8 +794,9 @@ function shellcheck_new_files() {
 }
 
 function latest_version() {
-  if [ $(current_branch) = "master" ]; then
-    # For master, simply use git tag without major version, this will work even if the release tag is not in the master
+  if [ $(current_branch) = "master" ] || [ $(current_branch) = "main" ]; then
+    # For main branch, simply use git tag without major version, this will work even
+    # if the release tag is not in the main
     git tag | sort -r --version-sort | head -n1
   else
     local semver=$(git describe --match "v[0-9]*" --abbrev=0)
--- a/library.sh
+++ b/library.sh
@@ -794,11 +794,16 @@ function shellcheck_new_files() {
 }
 
 function latest_version() {
-  local semver=$(git describe --match "v[0-9]*" --abbrev=0)
-  local major_minor=$(echo "$semver" | cut -d. -f1-2)
+  if [ $(current_branch) = "master" ]; then
+    # For master, simply use git tag without major version, this will work even if the release tag is not in the master
+    git tag | sort -r --version-sort | head -n1
+  else
+    local semver=$(git describe --match "v[0-9]*" --abbrev=0)
+    local major_minor=$(echo "$semver" | cut -d. -f1-2)
 
-  # Get the latest patch release for the major minor
-  git tag -l "${major_minor}*" | sort -r --version-sort | head -n1
+    # Get the latest patch release for the major minor
+    git tag -l "${major_minor}*" | sort -r --version-sort | head -n1
+  fi
 }
 
 # Initializations that depend on previous functions.
--- a/library.sh
+++ b/library.sh
@@ -794,6 +794,21 @@ function shellcheck_new_files() {
 }
 
 function latest_version() {
+  # This function works "best effort" and works on Prow but not necessarily locally.
+  # The problem is finding the latest release. If a release occurs on the same commit which
+  # was branched from master, then the tag will be an ancestor to any commit derived from master.
+  # That was the original logic. Additionally in a release branch, the tag is always an ancestor.
+  # However, if the release commit ends up not the first commit from master, then the tag is not
+  # an ancestor of master, so we can't use `git describe` to find the most recent versioned tag. So
+  # we just sort all the tags and find the newest versioned one.
+  # But when running locally, we cannot(?) know if the current branch is a fork of master or a fork
+  # of a release branch. That's where this function will malfunction when the last release did not
+  # occur on the first commit -- it will try to run the upgrade tests from an older version instead
+  # of the most recent release.
+  # Workarounds include:
+  # Tag the first commit of the release branch. Say release-0.75 released v0.75.0 from the second commit
+  # Then tag the first commit in common between master and release-0.75 with `v0.75`.
+  # Always name your local fork master or main.
   if [ $(current_branch) = "master" ] || [ $(current_branch) = "main" ]; then
     # For main branch, simply use git tag without major version, this will work even
     # if the release tag is not in the main
--- a/library.sh
+++ b/library.sh
@@ -797,7 +797,7 @@ function latest_version() {
   if [ $(current_branch) = "master" ] || [ $(current_branch) = "main" ]; then
     # For main branch, simply use git tag without major version, this will work even
     # if the release tag is not in the main
-    git tag | sort -r --version-sort | head -n1
+    git tag -l "v[0-9]*" | sort -r --version-sort | head -n1
   else
     local semver=$(git describe --match "v[0-9]*" --abbrev=0)
     local major_minor=$(echo "$semver" | cut -d. -f1-2)


library.sh
```diff
@@ -794,11 +794,16 @@ function shellcheck_new_files() {
 }
 
 function latest_version() {
-  local semver=$(git describe --match "v[0-9]*" --abbrev=0)
-  local major_minor=$(echo "$semver" | cut -d. -f1-2)
+  if [ $(current_branch) = "master" ]; then
```
vaikas, line 797: My git fu is not strong enough to reason about the actual PR, sorry... But the one thing we might want to do is include:
branch = "main"
here since we're moving away from using "master" so that would make this future proof.

zhongduo: Make sense. Fixed as suggested.

library.sh
```diff
@@ -794,11 +794,17 @@ function shellcheck_new_files() {
 }
 
 function latest_version() {
-  local semver=$(git describe --match "v[0-9]*" --abbrev=0)
-  local major_minor=$(echo "$semver" | cut -d. -f1-2)
+  if [ $(current_branch) = "master" ] || [ $(current_branch) = "main" ]; then
```
n3wscott, line 797: I would rather not hardcode these in, they are user set-able and programmatically discovery-able.

library.sh
```diff
@@ -794,11 +794,17 @@ function shellcheck_new_files() {
...
+    # For main branch, simply use git tag without major version, this will work even
+    # if the release tag is not in the main
+    git tag | sort -r --version-sort | head -n1
+  else
```
n3wscott, line 801: I think this breaks how the function is used.

n3wscott:
I do not understand what this PR is for, can you add detail to the description?

zhongduo:
> I do not understand what this PR is for, can you add detail to the description?

I have added #45 for explanation.

library.sh
zhongduo: It is fixing a problem that we are having in master/main branch. Other branches will stay the same.

library.sh
zhongduo: I didn't find a reliable way to know the default branch. How about just checking `! is_release`? I tend to minimize the impact, so specifying the branch names means less surprise to me, but I am OK with that too.

coryrc:
It's used here in serving:
https://github.com/knative/serving/blob/a39fbc0e2e75f7a7dd5129d98a23f09fd9f0d04c/test/e2e-common.sh#L58
then
https://github.com/knative/serving/blob/a39fbc0e2e75f7a7dd5129d98a23f09fd9f0d04c/test/e2e-common.sh#L219

which is currently run from master and usually has no problems. Is this going to change the behaviour?

zhongduo:
> It's used here in serving:
> https://github.com/knative/serving/blob/a39fbc0e2e75f7a7dd5129d98a23f09fd9f0d04c/test/e2e-common.sh#L58
> then
> https://github.com/knative/serving/blob/a39fbc0e2e75f7a7dd5129d98a23f09fd9f0d04c/test/e2e-common.sh#L219
> 
> which is currently run from master and usually has no problems. Is this going to change the behaviour?

As explained in the issue, most repos will work in master, problem will occur only when there is problem with the release process. This PR does not change the current behavior at all. It will keep getting the latest v0.20.x now from master. The original logic was more complex as it is need to work with all the release branches.

library.sh
coryrc: `! is_release_branch` should be okay

coryrc: (or better reverse the if/else, so `if is_release_branch; then`)

zhongduo: This might be a little bit tricky when running using a local branch that is targeting to be merged to a release branch. e.g. if you are working on a local branch based on v0.18, this latest_version will show v0.18 before, but now since the name of local branch is not a release branch, it will not use `git descirbe` so it will find 0.20 instead.

coryrc:
No commenting to the rescue! Glad these vague functions exist.

Are you saying the functionality of `latest_version` should be:

  Find me the latest tag whose major.minor is not greater than myself? (master being considered inf.inf)

If so, I'd contend that it's always been broken and still is. The solution here just papers over the problem of calculating the major.minor for a given commit. It depends on the commit that causes the first release to be the exact same as one in master, which, as seen here, is not always the case. To solve this without problems, you need to always tag 0.X.0 as a commit in master even if it isn't what you plan to release, but that also seems worse because what is a tag if not a release?

I think the compromise is to tag the first commit of the release branch with `0.20` -- no `.0`. Without a patch people shouldn't think it of a release but it should match the describe regex used by this function. The second half is to comment this function with a summary of my comment so we understand the limitations of this function.

zhongduo:
> Find me the latest tag whose major.minor is not greater than myself? (master being considered inf.inf)

Yes your understanding is correct.

> If so, I'd contend that it's always been broken and still is. The solution here just papers over the problem of calculating the major.minor for a given commit. It depends on the commit that causes the first release to be the exact same as one in master, which, as seen here, is not always the case. To solve this without problems, you need to always tag 0.X.0 as a commit in master even if it isn't what you plan to release, but that also seems worse because what is a tag if not a release?

I actually don't see any problem after this PR. IMO most repos have `git describe` works properly, and as you can see that's the case for most knative repos too. Now our automated Prow process does make this more difficult when a release process can fail and has to be fixed by new commit. However, this will only affect the master branch, as all the release branch is guaranteed to have the release tag in the branch. So after this PR the master branch will use `git tag` to search all the release tags instead of relying on `git describe` to work properly.

coryrc:
>  So after this PR the master branch will use `git tag` to search all the release tags instead of relying on `git describe` to work properly.
You point out elsewhere why you can't check if the branch is a release branch by looking at the branch name because PR forks can use a different branch. Doesn't that apply to checking for master too?

library.sh
coryrc: In this PR you are pushing from `zhongduo:fix-latest-version` to `knative:master`. This will be broken on your local copy, won't it?

coryrc:
I commented on the relevant code section re: master.

Won't tagging something `0.20` solve the problem though?

zhongduo:
> I commented on the relevant code section re: master.
> 
> Won't tagging something `0.20` solve the problem though?

Please refer to `eventing-kafka` again. The release tag `0.20` does already exist, but just that the commit with the tag is only on the release-0.20 branch, not master.

library.sh
zhongduo: I am not sure what you mean by "broken" here. But the problem in my previous comment will only happen when I am basing on a release branch if I change this PR to use `is_release_branch`.

coryrc:
> > Won't tagging something `0.20` solve the problem though?
> 
> Please refer to `eventing-kafka` again. The release tag `0.20` does already exist, but just that the commit with the tag is only on the release-0.20 branch, not master.

The tag `v0.20.0` exists and is not an ancestor of the tip of master but is an ancestor of the tip of release-0.20 branch. Agreed.

I am suggesting you tag `v0.20` (note: no last `.0`) to [the last common commit between release-0.20 and master](https://github.com/knative-sandbox/eventing-kafka/commit/6ea4099853e50372603a554d543ba89a73963b15). Note this tag isn't fully valid for releases, but is valid for the original regexes in this file:
```
local semver=$(git describe --match "v[0-9]*" --abbrev=0)
local major_minor=$(echo "$semver" | cut -d. -f1-2)
```
Thus, with the above tag added, `major_minor` will be `0.20` as we desire without having to change this file.

library.sh
coryrc: You don't want to use `is_release_branch` because is checks the branch name against a regex to see if it a release. This is valid point. But the current encoded logic is also to check against the branch name. Isn't this wrong for the same exact reason?

zhongduo:
> I am suggesting you tag `v0.20` (note: no last `.0`) to [the last common commit between release-0.20 and master](https://github.com/knative-sandbox/eventing-kafka/commit/6ea4099853e50372603a554d543ba89a73963b15). Note this tag isn't fully valid for releases, but is valid for the original regexes in this file:

I don't see why we would prefer such a complex solution to this PR. Note this rule has to be enforced to all the downstream repos such as `eventing-kafka`, not this repo. And it is also confusing to me to introduce such a tag that looks like release but not release just for this.

library.sh
zhongduo: Two reasons: 
1. minimal impact: `! is_release_branch` modifies the logic for all branches other than the ~20 release branches, while this PR only modify one branch.
2. For user forked branches with proper tag, using this PR it will continue working in this case. But with `! is_release_branch`, all the user local branches targeting release branches will fail.

coryrc: This solution won't work on forks of eventing-kafka master unless people's forked branch is named `master` or `main` until the next successful release.

zhongduo: That depends on how you run it. If it is a PR run by Prow it will work as Prow will test master with the PR merged. The only time it will fail is when you run the test locally. AFAIK, most repos have the tagging working properly, so I think the number of failure cases is reasonable.

coryrc: Then IMO trading one set of failures for a different set of failures + unknown possibly complications isn't worth just adding a single tag to eventing-kafka repo and documenting what to do for future release failures.

zhongduo: hmm, I am not sure if I will like having both v0.20 and v0.20.0 tags... But I will leave that to the community to make the call.

library.sh
```diff
@@ -794,11 +794,17 @@ function shellcheck_new_files() {
 }
 
 function latest_version() {
```
coryrc, line 796: Please add this comment and we can approve:

```
This function works "best effort" and works on Prow but not necessarily locally.
The problem is finding the latest release. If a release occurs on the same commit which
 was branched from master, then the tag will be an ancestor to any commit derived from master.
 That was the original logic. Additionally in a release branch, the tag is always an ancestor.
However, if the release commit ends up not the first commit from master, then the tag is not
 an ancestor of master, so we can't use `git describe` to find the most recent versioned tag. So
 we just sort all the tags and find the newest versioned one.
But when running locally, we cannot(?) know if the current branch is a fork of master or a fork
 of a release branch. That's where this function will malfunction when the last release did not
 occur on the first commit -- it will try to run the upgrade tests from an older version instead
 of the most recent release.
Workarounds include:
 Tag the first commit of the release branch. Say release-0.75 released v0.75.0 from the second commit
  Then tag the first commit in common between master and release-0.75 with `v0.75`.
 Always name your local fork master or main.
```

zhongduo: Added as suggested.

PR #44, opened 2021-01-14
Set GOBIN in codegen-library.sh

nak3:
If `GOBIN` is set to some directory, `go install` installs the binary
into GOBIN*1. And then, `hack/update-codegen.sh` (e.g. in serving repo)
fails due to missing `deepcopy-gen` binary installed by `go install` like:

```
--- Deepcopy Gen
hack/update-codegen.sh: line 84: /tmp/tmp.ZAaVX3NvIx/bin/deepcopy-gen: No such file or directory
```

Hence this PR sets `GOBIN` explicitly.

*1 https://golang.org/cmd/go/#hdr-Compile_and_install_packages_and_dependencies
> Executables are installed in the directory named by the GOBIN environment variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH environment variable is not set.

**Release Note**

```release-note
NONE
```

/cc @n3wscott

### DIFF
--- a/codegen-library.sh
+++ b/codegen-library.sh
@@ -26,6 +26,7 @@ else
 fi
 
 export GOPATH=$(go_mod_gopath_hack)
+export GOBIN=${GOPATH}/bin # Set GOBIN explicitly as deepcopy-gen is installed by go install.
 export MODULE_NAME=$(go_mod_module_name)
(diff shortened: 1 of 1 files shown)


n3wscott (commented):
/lgtm
/approve

PR #47, opened 2021-02-01
disable go's proxy and sumdb only for knative deps

dprotaso:
A git tag was re-written in some upstream dependency.  Thus causing `GOPROXY=direct` to fail due to difference hashes.

We actually have the ability to explicitly ski  the proxy and sumdb for only `knative.dev` repos - so lets do that.

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -539,7 +539,8 @@ function go_update_deps() {
 
   export GO111MODULE=on
   export GOFLAGS=""
-  export GOSUMDB=off   # Do not use the sum.golang.org service.
+  export GONOSUMDB="${GONOSUMDB},knative.dev/*"
+  export GONOPROXY="${GONOPROXY},knative.dev/*"
 
   echo "=== Update Deps for Golang"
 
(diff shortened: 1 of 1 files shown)


dprotaso:
cc @vaikas

vaikas:
@dprotaso is this the long term or short term? As in are we doing this for now to work around some issues, what should we do to prevent badness in the future for other dependencies?

library.sh
```diff
@@ -539,7 +539,8 @@ function go_update_deps() {
 
   export GO111MODULE=on
   export GOFLAGS=""
-  export GOSUMDB=off   # Do not use the sum.golang.org service.
+  export GONOSUMDB="${GONOSUMDB},knative.dev/*"
+  export GONOPROXY="${GONOPROXY},knative.dev/*"
```
n3wscott, line 543: is this the same as direct? Can you link to a doc that describes this behaviour?

dprotaso: https://golang.org/cmd/go/#hdr-Module_configuration_for_non_public_modules

mattmoor (approved):

PR #51, opened 2021-03-03
Schema tool for reflective schema generation.

n3wscott:
# Changes

- :gift: New schema generator based on reflection and go comments. 

/kind  enhancement

I found an interesting tool that @Harwayne built inside of Knative-GCP, I have cherrypicked that tool and created the start of a CLI from that code. Work in progress.

In use https://github.com/knative-sandbox/discovery/pull/153

**Release Note**

```release-note
Adding the start of a base cli for generating schema examples in place for downstream projects.
```


n3wscott:
/approve

we are going to start with this to get to 1.0 and then we can move to something better. This gets us close without needing to do a ton of manual labor.

PR #54, opened 2021-03-12
go test flags should go first

dprotaso:
Should allow downstream to append `-args`

https://golang.org/cmd/go/#hdr-Testing_flags

> To keep an argument for a test binary from being interpreted as a known flag or a package name, use -args (see 'go help test') which passes the remainder of the command line through to the test binary uninterpreted and unaltered.
> 
> For instance, the command
> 
> ```
> go test -v -args -x -v
> ```
> will compile the test binary and then run it as
> 
> ```
> pkg.test -test.v -x -v
> ```


n3wscott (approved):
/lgtm

PR #55, opened 2021-03-15
Don't redirect STDERR to /dev/null when querying pods

arturenault:
This function checks the status of pods in a namespace. It redirects STDERR to /dev/null, which makes failures hard to debug. This redirection is unnecessary as STDERR gets logged to console when using command substitution anyways. Not redirecting it would make failures easier to debug.


n3wscott (approved):
/lgtm

We might revert this if we find the logs getting huge?

PR #67, opened 2021-04-27
Fix default codegen pkg dir

benmoss:
When running in CI the directory isn't always called 'pkg'

# Changes
- :bug: Don't assume the pkg dir is named `pkg` in the codegen lib script

/kind bug

We just started using this script in knative/pkg and it looks like it doesn't work in this edge-case scenario where pkg is the repo being updated, but it's checked out as `/home/runner/work/knobots/knobots` on Github Actions.

https://github.com/knative-sandbox/knobots/runs/2448216473?check_suite_focus=true

/assign @n3wscott @dprotaso

### DIFF
--- a/codegen-library.sh
+++ b/codegen-library.sh
@@ -29,7 +29,7 @@ export GOPATH=$(go_mod_gopath_hack)
 export GOBIN=${GOPATH}/bin # Set GOBIN explicitly as deepcopy-gen is installed by go install.
 export MODULE_NAME=$(go_mod_module_name)
 export CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${REPO_ROOT_DIR}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
-export KNATIVE_CODEGEN_PKG=${KNATIVE_CODEGEN_PKG:-$(cd ${REPO_ROOT_DIR}; ls -d -1 ./vendor/knative.dev/pkg 2>/dev/null || echo ../pkg)}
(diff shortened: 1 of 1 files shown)


codegen-library.sh
```diff
@@ -29,7 +29,7 @@ export GOPATH=$(go_mod_gopath_hack)
 export GOBIN=${GOPATH}/bin # Set GOBIN explicitly as deepcopy-gen is installed by go install.
 export MODULE_NAME=$(go_mod_module_name)
 export CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${REPO_ROOT_DIR}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
-export KNATIVE_CODEGEN_PKG=${KNATIVE_CODEGEN_PKG:-$(cd ${REPO_ROOT_DIR}; ls -d -1 ./vendor/knative.dev/pkg 2>/dev/null || echo ../pkg)}
+export KNATIVE_CODEGEN_PKG=${KNATIVE_CODEGEN_PKG:-$(cd ${REPO_ROOT_DIR}; ls -d -1 ./vendor/knative.dev/pkg 2>/dev/null || echo "${REPO_ROOT_DIR}")}
```
n3wscott, line 32: I think this will work for the `pkg` repo but seems like it will fail for all other repos. TBH, i don't understand when it is the case where vendor does not have pkg... likely some out of tree integration? pre-gomod?

Given that all other repos in Knative should vendor, I think this is a safe change, and if it breaks someone out of tree they can help make a solution that works for them and us. `../pkg` is not a safe assumption in a go mod world.

/lgtm
/approve

n3wscott (approved):

PR #72, opened 2021-06-08
Skip GOPATH hacks if the repository is already on GOPATH

markusthoemmes:
# Changes

As per title, this supports the case where the repositories are still checked out onto a working `$GOPATH`.

1. Moves the module check above the GOPATH hackery. This took ages to fetch if running in a `$GOPATH` scenario, for whatever reason.
2. Make the GOPATH hackery skip itself if the repository is already checked out onto a legit `$GOPATH`. It's not necessary then.
3. Create a random directory for `$GOBIN` to not pollute the local bin directory even if running on a legit `$GOPATH`.

/assign @n3wscott @slinkydeveloper

### DIFF
--- a/codegen-library.sh
+++ b/codegen-library.sh
@@ -25,9 +25,9 @@ else
   exit
 fi
 
-export GOPATH=$(go_mod_gopath_hack)
-export GOBIN=${GOPATH}/bin # Set GOBIN explicitly as deepcopy-gen is installed by go install.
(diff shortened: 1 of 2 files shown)


slinkydeveloper:
This works very well for me, it solved my issues with veeeeery long codegen script execution :heart_eyes: , thanks!

n3wscott (approved):
/lgtm

Sorry it was not working with it checked out in GOPATH, odd that prow never complained.

PR #73, opened 2021-06-09
Rollback random GOBIN and make it relative to GOPATH again

markusthoemmes:
# Changes

This should fix the current breakage in our nightly update scripts. Tested this in an empty docker container (as opposed to my system) to make sure it actually works.

### DIFF
--- a/codegen-library.sh
+++ b/codegen-library.sh
@@ -27,7 +27,7 @@ fi
 
 export MODULE_NAME=$(go_mod_module_name)
 export GOPATH=$(go_mod_gopath_hack)
-export GOBIN="$(mktemp -d)" # Set GOBIN explicitly as deepcopy-gen is installed by go install.
+export GOBIN=${GOPATH}/bin # Set GOBIN explicitly as deepcopy-gen is installed by go install.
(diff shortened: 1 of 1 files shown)


codegen-library.sh
```diff
@@ -27,7 +27,7 @@ fi
 
 export MODULE_NAME=$(go_mod_module_name)
 export GOPATH=$(go_mod_gopath_hack)
-export GOBIN="$(mktemp -d)" # Set GOBIN explicitly as deepcopy-gen is installed by go install.
+export GOBIN=${GOPATH}/bin # Set GOBIN explicitly as deepcopy-gen is installed by go install.
```
julz, line 30: Just to reflect some context from slack to here, this does mean that we'll pollute the real GOPATH/bin in the $GOPATH case (but not the go mod case). I think that's fine, but I'm biased since I don't use $GOPATH any more ;).

markusthoemmes: As one of the lone GOPATH users, I think it's fine :joy:

julz (approved):
/lgtm

markusthoemmes:
/retest

unrelated.

PR #77, opened 2021-08-05
Stop performing pre-releases

evankanderson:
Fixes https://github.com/knative/client/issues/1404

If we decide we want to make this configurable on a per-repo basis, it's pretty easy to go back and mark existing releases as `prerelease`; see https://github.com/knative/client/issues/1404#issuecomment-893754761 for an example script.


rhuss (commented):
/approve
/lgtm

Issue #97, opened 2021-11-01
Make sure Prow Bot doesn't bump Go version to 1.0 due to automation 


omerbensaadon:
See: https://github.com/knative/hack/blob/a42c72a8fc009dc5b0878fa3c8143e15b07b851b/release.sh#L180-L193 

We want to make sure that when the release gets tagged, prow bot bumps the GO version to 0.27 not 1.0. 

See Julz's [Release-1.0 machinery changes](https://docs.google.com/document/d/1_khEw5_l8Zhh4FuY3NM56o_OM-cNWmmq1vXfLCxZqMA/edit) for context

- Need to cherry-pick this change to the 1.0 branch in `hack` 
- Run the knobots so this change gets propagated throughout
- There may be a prow config update?

PR #98, opened 2021-11-01
Release use `knative-vX.Y.Z` tag and infer go module tag version automatically

dprotaso:
Fixes: https://github.com/knative/hack/issues/97

This updates the release scripts to support two tag schemes:
- knative-vX.Y.Z
- vX.Y.Z

Note: when publishing to GitHub we look at the desired tag and if it matches `v1*` then we'll push the corresponding go module version tag.

The formula for the go module version is
- drop the major to 0
- add 27 to the minor
- patch version remains the same

See the document here: https://docs.google.com/document/d/1_khEw5_l8Zhh4FuY3NM56o_OM-cNWmmq1vXfLCxZqMA/edit?usp=sharing

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -135,8 +135,7 @@ function master_version() {
 # For example, "v0.2.1" returns "2"
 # Parameters: $1 - release version label.
 function minor_version() {
-  local release="${1//v/}"
-  local tokens=(${release//\./ })
+  local tokens=(${1//\./ })
   echo "${tokens[1]}"
 }
 
@@ -631,6 +630,8 @@ function publish_to_github() {
   # v1.0.0 = v0.27.0
   # v1.0.1 = v0.27.1
   # v1.1.1 = v0.28.1
+  #
+  # See: https://github.com/knative/hack/pull/97
   if [[ "$TAG" == "v1"* ]]; then
     local release_minor=$(minor_version $TAG)
     local go_module_version="v0.$(( release_minor + 27 )).$(patch_version $TAG)"
--- a/release.sh
+++ b/release.sh
@@ -131,10 +131,19 @@ function master_version() {
   echo "${tokens[0]}.${tokens[1]}"
 }
 
+# Return the minor version of a release.
+# For example, "v0.2.1" returns "2"
+# Parameters: $1 - release version label.
+function minor_version() {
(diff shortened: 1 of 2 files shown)


chaodaiG:
I haven't followed this for quite a while, unassign for now
/unassign
please add me back if just need a rubberstamp

release.sh
```diff
@@ -604,16 +623,28 @@ function publish_to_github() {
...
+
+  # Include a tag for the go module version
+  #
+  # v1.0.0 = v0.27.0
+  # v1.0.1 = v0.27.1
+  # v1.1.1 = v0.28.1
+  if [[ "$TAG" == "v1"* ]]; then
+    local release_minor=$(minor_version $TAG)
+    local go_module_version="0.$(( release_minor + 27 )).$(patch_version $TAG)"
+    git tag -a "${go_module_version}" -m "${title}"
```
benmoss, line 637: need a `v` prefix in here somewhere

dprotaso: added to the declaration of `go_module_version` variable

release.sh
```diff
@@ -131,10 +131,19 @@ function master_version() {
   echo "${tokens[0]}.${tokens[1]}"
 }
 
+# Return the minor version of a release.
+# For example, "v0.2.1" returns "2"
+# Parameters: $1 - release version label.
+function minor_version() {
+  local release="${1//v/}"
+  local tokens=(${release//\./ })
```
chizhg, line 139: Can we keep the logic consistent with the `patch_version` function by splitting with `.`?

release.sh
```diff
@@ -604,16 +623,28 @@ function publish_to_github() {
   if [[ -n "${RELEASE_NOTES}" ]]; then
     cat "${RELEASE_NOTES}" >> "${description}"
   fi
-  git tag -a "${TAG}" -m "${title}"
-  git_push tag "${TAG}"
+  git tag -a "${github_tag}" -m "${title}"
+  git_push tag "${github_tag}"
+
+  # Include a tag for the go module version
```
chizhg, line 629: Can we add the link for the issue / Slack thread so people can get more contexts on how the decision was made?

chizhg:
/lgtm

/hold
In case Ben has other comments.

PR #101, opened 2021-11-02
fix the latest_version function

dprotaso:
it's aware of the new tag format and the manual jump from
v0.26 => v1.0

note: lates_version returns the tag which may now include prefixes

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -64,6 +64,31 @@ fi
 # On a Prow job, redirect stderr to stdout so it's synchronously added to log
 (( IS_PROW )) && exec 2>&1
 
+# Return the major version of a release.
+# For example, "v0.2.1" returns "0"
+# Parameters: $1 - release version label.
+function major_version() {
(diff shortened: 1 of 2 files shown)


dprotaso:
verified the behaviour 

- serving, main => picked knative-v1.0.0
- serving, release-1.0 => picked v0.26.0

tzununbekov:
I was wondering if using Github's `latest` endpoint could help save some lines:
```bash
function latest_version() {
  local origin=$(git config --get remote.origin.url | sed 's/.*:\(.*\)\..*$/\1/')
  curl -I -s https://github.com/${origin}/releases/latest | grep "location:" | awk -F/ '{print $NF}' | sed 's/ //g'
}
```

dprotaso:
looking https://docs.github.com/en/rest/reference/repos#get-the-latest-release - it's is sorted by `created_at` time 

but we need to sort by our semver versions since we'll produce patch releases for older branches

PR #104, opened 2021-11-10
:bug: Allows to see TAG from build_release func

cardil:
# Changes

- :bug: Allows to see `TAG` and `TAG_RELEASE` from build_release func

/kind bug

Fixes https://github.com/knative-sandbox/kn-plugin-event/issues/77

## Justification

In https://github.com/knative-sandbox/kn-plugin-event/pull/90 we had implemented a way to properly tag binaries and images from values of `$TAG` (only when `TAG_RELEASE==1`). Unfortunately, those variables aren't exported, so the build process can't see them.

This change exports those two variables, so the release script will work properly for kn event plugin.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -82,10 +82,10 @@ function publish_to_gcs() {
 # These are global environment variables.
 SKIP_TESTS=0
 PRESUBMIT_TEST_FAIL_FAST=1
-TAG_RELEASE=0
+export TAG_RELEASE=0
 PUBLISH_RELEASE=0
 PUBLISH_TO_GITHUB=0
-TAG=""
+export TAG=""
 BUILD_COMMIT_HASH=""
 BUILD_YYYYMMDD=""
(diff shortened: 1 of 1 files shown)


dsimansk:
@dprotaso @chizhg can you please take a look?

release.sh
```diff
@@ -104,6 +104,7 @@ export KO_DOCKER_REPO="gcr.io/knative-nightly"
 # Build stripped binary to reduce size
 export GOFLAGS="-ldflags=-s -ldflags=-w"
 export GITHUB_TOKEN=""
+export TAG TAG_RELEASE
```
dprotaso, line 107: Can you just add `export` to line 85 and 88?

cardil: In this case, yes, I can.

I often write it like that to not fall onto [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155).

PR #105, opened 2021-11-11
chore: further explain knative CRD-generation support

dan-j:
Many new contributors will have had experience with other k8s tooling like the Operator SDK and as
such may expect CRD generation to be a standard feature within the Knative tooling (myself being
an example). This change adds an extra explanation of Knative's position on CRD generation, and 
should make this tool more easily-discoverable via GitHub search by including keywords like
"CRD generation" and "controller-gen".

# Changes

:broom: Update README.md

/kind documentation

This isn't assigned an issue, I needed to ask on Slack to find this repo so thought it would be worth clarifying for future new contributors.

### DIFF
--- a/schema/README.md
+++ b/schema/README.md
@@ -2,7 +2,11 @@
 
 Schema is a seed of a CLI tool that a downstream can use to use reflection and go file inspection to 
 generate a base version of OpenAPI for a CRD. The resulting schema will be used by Kubernetes to
-provide results from `kubectl explain <type>` calls, and type validation.  
+provide results from `kubectl explain <type>` calls, and type validation.
+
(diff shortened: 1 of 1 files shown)


dprotaso:
/ok-to-test
/lgtm
/approve

Just an FYI - for serving we are using a custom fork of controller-gen so that we can make modifications to an embedded PodSpec ie. https://github.com/knative/serving/blob/main/hack/update-schemas.sh 

I believe some knative eventing repos are using this tool to just bootstrap the schema initially.

I think ideally we'd want to use controller-gen but need someone to drive some changes upstream and across the Knative project.

PR #127, opened 2021-12-06
[release-0.26] Fix the latest_version function (#101)

pierDipi:
Cherry-pick of: https://github.com/knative/hack/pull/101

/cc @dprotaso

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -832,29 +832,26 @@ function shellcheck_new_files() {
 function latest_version() {
   local branch_name="$(current_branch)"
 
-  # Use the latest release for main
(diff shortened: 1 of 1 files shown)


dprotaso:
#101 was actually broken and had subsequent follow up PRs


What's in release-0.26 is the same as release-1.0 and HEAD. So this PR is reverting to a buggier version. What issue are you seeing?

ie. https://github.com/knative/hack/blob/e11ac125e70709d1beec59d81116b392193c01fb/library.sh#L865-L915

pierDipi:
Thanks!

Issue #136, opened 2022-01-10
`e2e-tests.sh` fails if teardown function isn't defined


cardil:
https://github.com/knative/hack/blob/abd085fc43de80d6344a88788ae6b926c48c908e/e2e-tests.sh#L31-L33

If `test_teardown` and `knative_teardown` isn't defined, the script exits with error instead of passing properly.

/kind bug

cardil:
Affects https://github.com/knative-sandbox/reconciler-test/pull/273

PR #137, opened 2022-01-10
:bug: No longer fail if teardown function isn't defined

cardil:
# Changes

- :bug: No longer fail if teardown function isn't defined knative/hack#136

/kind bug

Fixes knative/hack#136

### DIFF
--- a/e2e-tests.sh
+++ b/e2e-tests.sh
@@ -17,7 +17,7 @@
 # This is a helper script for Knative E2E test scripts.
 # See README.md for instructions on how to use it.
 
-source $(dirname "${BASH_SOURCE[0]}")/infra-library.sh
+source "$(dirname "${BASH_SOURCE[0]:-$0}")/infra-library.sh"
 
 readonly TEST_RESULT_FILE=/tmp/${REPO_NAME}-e2e-result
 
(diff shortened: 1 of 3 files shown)


e2e-tests.sh
```diff
@@ -29,8 +29,12 @@ function teardown_test_resources() {
   # On boskos, save time and don't teardown as the cluster will be destroyed anyway.
   (( IS_BOSKOS )) && return
   header "Tearing down test environment"
-  function_exists test_teardown && test_teardown
-  function_exists knative_teardown && knative_teardown
+  if function_exists test_teardown; then
```
chizhg, line 32: Based on my knowledge on Bash, this should be the same as before.
Could you elaborate what was the bug for L32-33 and how this change fixes it?

cardil: Oh, that's very different.

```bash
#!/bin/bash

set -Eeuo pipefail

function function_exists() {
  [[ "$(type -t $1)" == "function" ]]
}

function_exists test_teardown && test_teardown

echo "Result: $?"
```

You will get result: `Result: 1`. That's because line `function_exists test_teardown && test_teardown` returns 1. And that's last execution.

When you use the following. The `function_exists test_teardown` is executed as a boolean expression, and the if returns 0, even if the condition is not met (that's the `fi` command BTW).

```bash
if function_exists test_teardown; then
  test_teardown
fi
```

chizhg: Ahh, that's a huge Bash pitfall..
Thanks for explaining!

cardil: There is plenty of those. Bash only seams obvious.

PR #138, opened 2022-01-13
Disable Logs/Metrics on GKE and Run Preemptible Nodes

upodroid:
Part of https://github.com/knative/test-infra/issues/3003

### DIFF
--- a/infra-library.sh
+++ b/infra-library.sh
@@ -137,5 +137,6 @@ function create_gke_test_cluster() {
   local -n _custom_flags=$1
   local -n _test_command=$2
 
-  run_kntest kubetest2 gke "${_custom_flags[@]}" --test-command="${_test_command[*]}"
+# We are disabling logs and metrics on Boskos Clusters as they are not used.
+  run_kntest kubetest2 gke "${_custom_flags[@]}" --test-command="${_test_command[*]}" --extra-gcloud-flags="--logging=NONE --metrics=NONE --preemptible"
 }
--- a/infra-library.sh
+++ b/infra-library.sh
@@ -137,9 +137,8 @@ function create_gke_test_cluster() {
   local -n _custom_flags=$1
   local -n _test_command=$2
 
-# We are disabling logs and metrics on Boskos Clusters by default as they are not used. Manually set GKE_DISABLE_TELEMETRY to false to enable logging
-  if [ "$GKE_DISABLE_TELEMETRY" == "false" ]
-  then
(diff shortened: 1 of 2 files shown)


chizhg:
> Can we make this configurable? For example when debugging a flake I'd like to opt into logging to do better searches

+1 on this.

@upodroid the image with the updated gcloud version has been rolled out, could you make this change?

upodroid:
I added an env variable that enables logging when needed

infra-library.sh
```diff
@@ -137,5 +137,11 @@ function create_gke_test_cluster() {
   local -n _custom_flags=$1
   local -n _test_command=$2
 
-  run_kntest kubetest2 gke "${_custom_flags[@]}" --test-command="${_test_command[*]}"
+# We are disabling logs and metrics on Boskos Clusters by default as they are not used. Manually set GKE_DISABLE_TELEMETRY to false to enable logging
```
chizhg, line 140: Super nit: the indentation is not correct.

infra-library.sh
```diff
@@ -137,5 +137,11 @@ function create_gke_test_cluster() {
...
+  if [ "$GKE_DISABLE_TELEMETRY" == "false" ]
```
chizhg, line 141: Please keep the Bash style consistent across the scripts to change it to `if [[ "$GKE_DISABLE_TELEMETRY" == "false" ]]; then`.

Also it's more natural to set default value as false, so I would recommend something like `"${GKE_ENABLE_TELEMETRY:=false}"`

chizhg (commented):
Left some nits. Also please fix the unit tests.

infra-library.sh
upodroid: fixed

infra-library.sh
upodroid: fixed

infra-library.sh
```diff
@@ -137,5 +137,10 @@ function create_gke_test_cluster() {
   local -n _custom_flags=$1
   local -n _test_command=$2
 
-  run_kntest kubetest2 gke "${_custom_flags[@]}" --test-command="${_test_command[*]}"
+  # We are disabling logs and metrics on Boskos Clusters by default as they are not used. Manually set ENABLE_GKE_TELEMETRY to true to enable logging
+  if [[ "$ENABLE_GKE_TELEMETRY" == "true" ]]; then
+    run_kntest kubetest2 gke "${_custom_flags[@]}" --test-command="${_test_command[*]}" --extra-gcloud-flags="--preemptible"
```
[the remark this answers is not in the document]
dprotaso: cc @chizhg @upodroid

PR #144, opened 2022-01-24
Enable opting out Preemptible/Spot GKE Worker Nodes

upodroid:
/cc @dprotaso @chizhg﻿

related to #138

### DIFF
--- a/infra-library.sh
+++ b/infra-library.sh
@@ -123,9 +123,14 @@ function create_gke_test_cluster() {
   local -n _custom_flags=$1
   local -n _test_command=$2
 
-  # We are disabling logs and metrics on Boskos Clusters by default as they are not used. Manually set ENABLE_GKE_TELEMETRY to true to enable logging
-  if [[ "$ENABLE_GKE_TELEMETRY" == "true" ]]; then
+  # We are disabling logs and metrics on Boskos Clusters by default as they are not used. Manually set ENABLE_GKE_TELEMETRY to true to enable telemetry
+  # and DISABLE_PREEMPTIBLE_NODES to true to create normal VMs. VM Preemption is a rare event and shouldn't be distruptive given the fault tolerant nature of our tests.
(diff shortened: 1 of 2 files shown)


infra-library.sh
```diff
@@ -123,9 +123,14 @@ function create_gke_test_cluster() {
   local -n _custom_flags=$1
   local -n _test_command=$2
 
-  # We are disabling logs and metrics on Boskos Clusters by default as they are not used. Manually set ENABLE_GKE_TELEMETRY to true to enable logging
-  if [[ "$ENABLE_GKE_TELEMETRY" == "true" ]]; then
+  # We are disabling logs and metrics on Boskos Clusters by default as they are not used. Manually set ENABLE_GKE_TELEMETRY to true to enable telemetry
```
chizhg, line 126: The `if...else` here is a bit hard to read. I would recommend changing them to something like below:
```bash
local extra_gcloud_flags=""
if [[ "$ENABLE_GKE_TELEMETRY" == "true" ]]
  extra_gcloud_flags="${extra_gcloud_flags} --logging=NONE --monitoring=NONE"
fi
if [[ "$DISABLE_PREEMPTIBLE_NODES" == "true" ]]
  extra_gcloud_flags="${extra_gcloud_flags} --preemptible"
fi
run_kntest kubetest2 gke "${_custom_flags[@]}" --test-command="${_test_command[*]}" --extra-gcloud-flags="${extra_gcloud_flags}"
```

dprotaso:
sorry - can we make `preemptible` opt-in?

infra-library.sh
upodroid: Fixed, @dprotaso I made the preemptible opt-in

chizhg:
Would be better to add this in the README, more specifically under `notes` in the `## Using the `e2e-tests.sh` helper script` section. In the future we need a better place for referencing these hidden env vars.

/lgtm

upodroid:
I'll do that tomorrow

PR #169, opened 2022-04-12
only use `install` for run_go_tool

kvmware:
# Changes

- :broom: Use `go install` instead of `get` for `run_go_tool` function.
- :wastebasket: remove the root check logic as `go install` does not change the `go.mod` file.

/kind cleanup

Fixes # https://github.com/knative/test-infra/issues/3113

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -544,26 +544,19 @@ function start_latest_eventing_sugar_controller() {
 }
 
 # Run a go tool, installing it first if necessary.
-# Parameters: $1 - tool package/dir for go get/install.
+# Parameters: $1 - tool package/dir for go install.
 #             $2 - tool to run.
 #             $3..$n - parameters passed to the tool.
 function run_go_tool() {
+  local package=$1
   local tool=$2
   local install_failed=0
+  # If no `@version` is provided, default to adding `@latest`
+  if [[ "$package" != *@* ]]; then
+    package=$package@latest
+  fi
   if [[ -z "$(which ${tool})" ]]; then
-    local action=get
-    [[ $1 =~ ^[\./].* ]] && action=install
-    # Avoid running `go get` from root dir of the repository, as it can change go.sum and go.mod files.
-    # See discussions in https://github.com/golang/go/issues/27643.
-    if [[ ${action} == "get" && $(pwd) == "${REPO_ROOT_DIR}" ]]; then
-      local temp_dir="$(mktemp -d)"
-      # Swallow the output as we are returning the stdout in the end.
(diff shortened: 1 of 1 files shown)


library.sh
```diff
@@ -544,26 +544,15 @@ function start_latest_eventing_sugar_controller() {
...
-      local temp_dir="$(mktemp -d)"
-      # Swallow the output as we are returning the stdout in the end.
-      pushd "${temp_dir}" > /dev/null 2>&1
-      GOFLAGS="" go ${action} "$1" || install_failed=1
-      popd > /dev/null 2>&1
-    else
-      GOFLAGS="" go ${action} "$1" || install_failed=1
-    fi
+    GOFLAGS="" go install "$1" || install_failed=1
+    echo $install_failed $1
```
chizhg, line 555: Is this line needed or only for debugging?

kvmware: Ah, good catch. removed.

chizhg:
```
2022/04/12 23:55:54 Failed to get k8s version kubernetes version "1.21.1" is not compatible, need at least "1.22.0-0" (this can be overridden with the env var "KUBERNETES_MIN_VERSION")
```

It seems we need to update the version of `kind` CLI. Or if no one would want to create Prow jobs run with `kind` clusters, we can drop this Prow job.

chizhg:
> /retest

You'd need to wait for a PR like https://github.com/knative/test-infra/pull/3261 to bump the image version, and then /retest.

cardil:
I think that instead of doing the `install` and the run, we could leverage Go 1.17+ `go run package@version` construct. That will not modifying the user's PC.

library.sh
```diff
@@ -544,26 +544,19 @@ function start_latest_eventing_sugar_controller() {
...
-    if [[ ${action} == "get" && $(pwd) == "${REPO_ROOT_DIR}" ]]; then
-      local temp_dir="$(mktemp -d)"
-      # Swallow the output as we are returning the stdout in the end.
-      pushd "${temp_dir}" > /dev/null 2>&1
-      GOFLAGS="" go ${action} "$1" || install_failed=1
-      popd > /dev/null 2>&1
-    else
-      GOFLAGS="" go ${action} "$1" || install_failed=1
-    fi
+    GOFLAGS="" go install "$package" || install_failed=1
```
dprotaso, line 559: nit: for follow up - unsure if we want to drop `GOFLAGS=""`

upodroid:
/cherrypick release-1.2

@haubenr

PR #172, opened 2022-04-27
:broom: Migrate to using the `go run` instead of `go install`

### LINKED ISSUE #170: Consider using `go run` instead of `go install` then `run` in library.sh
Currently `function run_go_tool()` in `library.sh` first installs the tool (without polluting go.mod), then runs the tool. @cardil suggests we use `go run` instead.

cardil:
# Changes

- :broom: Migrate to using the `go run` instead of `go install`
- :gift: Collecting JSONL, ANSI and nice HTML logs from all go tests
- :wastebasket: Removal of redundant manual race detection processing, and empty test suites handlers

/kind cleanup

Fixes #170

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -40,7 +40,7 @@ fi
 readonly IS_PROW
 [[ ! -v REPO_ROOT_DIR ]] && REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
 readonly REPO_ROOT_DIR
-readonly REPO_NAME="$(basename ${REPO_ROOT_DIR})"
+readonly REPO_NAME="${REPO_NAME:-$(basename "${REPO_ROOT_DIR}")}"
 
 # Useful flags about the current OS
 IS_LINUX=0
@@ -56,10 +56,15 @@ readonly IS_LINUX
 readonly IS_OSX
 readonly IS_WINDOWS
 
+export TMPDIR="${TMPDIR:-$(mktemp -u -t -d knative.XXXXXXXX)}"
+mkdir -p "$TMPDIR"
 # Set ARTIFACTS to an empty temp dir if unset
 if [[ -z "${ARTIFACTS:-}" ]]; then
-  export ARTIFACTS="$(mktemp -d)"
+  ARTIFACTS="$(mktemp -u -t -d)"
+  export ARTIFACTS
 fi
+mkdir -p "$ARTIFACTS"
+
 
 # On a Prow job, redirect stderr to stdout so it's synchronously added to log
 (( IS_PROW )) && exec 2>&1
@@ -429,7 +434,7 @@ function mktemp_with_extension() {
 function create_junit_xml() {
   local xml
   xml="$(mktemp_with_extension "${ARTIFACTS}"/junit_XXXXXXXX xml)"
-  echo "JUnit file ${xml} is created for reporting the test result"
+  echo "XML report for $1::$2 written to ${xml}"
   run_kntest junit --suite="$1" --name="$2" --err-msg="$3" --dest="${xml}" || return 1
 }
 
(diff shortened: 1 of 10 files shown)


library.sh
```diff
@@ -429,45 +434,46 @@ function mktemp_with_extension() {
...
+    -showteststatus \
+    -nofail > "$ansilog"
+  echo "Test log (ANSI) written to ${ansilog}"
+
+  htmllog="${logfile/.jsonl/.html}"
+  go_run github.com/buildkite/terminal-to-html/v3/cmd/terminal-to-html@v3.6.1 \
+    --preview < "$ansilog" > "$htmllog"
+  echo "Test log (HTML) written to ${htmllog}"
+
+  return ${gotest_retcode}
```
chizhg, line 476: Why collecting these?

cardil: We are running tests without the flag `-v`. That causes the passing tests to have no output. That makes perfect sense. Test passing, so why bother looking at logs, right?

But, sometimes there is a need to see how exactly the test passed. That necessity is often present for e2e tests, but maybe not only for them. The full output is handy, then. 

The _JSONL_ contains all lines of test output, see [example log](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_hack/172/unit-tests_hack_main/1519415863650816000/artifacts/go_test_xUZhYeE4.jsonl). The ANSI and HTML versions are just for human continence, as they are much clearer to read - see [HTML example](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_hack/172/unit-tests_hack_main/1519415863650816000/artifacts/go_test_xUZhYeE4.html).

library.sh
```diff
@@ -429,45 +434,46 @@ function mktemp_with_extension() {
...
+  logfile="${logfile/.xml/.jsonl}"
   echo "Running go test with args: ${go_test_args[*]}"
-  capture_output "${report}" gotestsum --format "${GO_TEST_VERBOSITY:-testname}" \
-    --junitfile "${xml}" --junitfile-testsuite-name relative --junitfile-testcase-classname relative \
+  go_run gotest.tools/gotestsum@v1.8.0 \
+    --format "${GO_TEST_VERBOSITY:-testname}" \
+    --junitfile "${xml}" \
+    --junitfile-testsuite-name relative \
+    --junitfile-testcase-classname relative \
+    --jsonfile "${logfile}" \
```
evankanderson, line 456: Do the `junitfile` and `jsonfile` have different contents? If not, I'd prefer to collect one (whichever has more information), and then convert from one to the other later if necessary. Cloud storage may be cheap, but it's not completely free, and keeping two copies of stuff that we may never look at seems excessive.

library.sh
```diff
@@ -429,45 +434,46 @@ function mktemp_with_extension() {
...
+  go_run github.com/haveyoudebuggedit/gotestfmt/v2/cmd/gotestfmt@v2.3.1 \
+    -input "${logfile}" \
+    -showteststatus \
+    -nofail > "$ansilog"
+  echo "Test log (ANSI) written to ${ansilog}"
+
+  htmllog="${logfile/.jsonl/.html}"
+  go_run github.com/buildkite/terminal-to-html/v3/cmd/terminal-to-html@v3.6.1 \
+    --preview < "$ansilog" > "$htmllog"
+  echo "Test log (HTML) written to ${htmllog}"
```
evankanderson, line 474: Rather than running these every time, how about providing instructions for users on how to do these if needed?

gab-satchi:
Hey @cardil I'm working on the 1.5 release, is this something that can wait for after 1.5?

dprotaso:
@gab-satchi this PR isn't material to the 1.5 release so we're good to cut

library.sh
cardil: Yes, they differ.

The junit file contains the XML with test names and execution time. If test is successful, this XML doesn't contain the test output if the test is successful. See, [example file](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_hack/172/unit-tests_hack_main/1519415863650816000/artifacts/junit_xUZhYeE4.xml). This file is read by Prow as well.

The json file contains a JSON object for every line of output the tests have produced, including the Golang output and the output of the successful tests as well. This can be handy for integration tests and upgrade tests. For both of them, the real reason for failure in some step might be hidden in the output of previous successful `t.Run` subtest - for example, an WARN log message. The idea is to archive both of them, in case the debugging is needed.

library.sh
cardil: The original JSON file isn't easy to understand for humans - see [example file](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_hack/172/unit-tests_hack_main/1519415863650816000/artifacts/go_test_xUZhYeE4.jsonl). Providing an easy-to-read HTML log file would greatly enhance the usability of test reports - see [example file](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_hack/172/unit-tests_hack_main/1519415863650816000/artifacts/go_test_xUZhYeE4.html)

I think most of the people will not know how to use the original JSON lines report file, and I don't see a place to post instructions on how to read it. The additional cost of creating the HTML report and keeping it on cloud storage isn't large. This is just a test output, kilobytes, even in the worst cases.

evankanderson:
Assuming this doesn't increase our storage costs too much, I'm happy to review / approve this if it's still wanted (was clearing out my notifications).

cardil:
@evankanderson yes, please.

cardil:
@evankanderson Would you mind slapping a `/lgtm`?

kvmware:
LGTM @evankanderson knows golang way better than I so he should probably /LGTM it

evankanderson (commented):
/lgmt
/approve

PR #173, opened 2022-04-29
Skip passing enable-ha flag to test script

psschwei:
Signed-off-by: Paul S. Schweigert <paulschw@us.ibm.com>

Serving uses the `--enable-ha` flag in several of its e2e tests, but
parsing it out causes the flag to end up getting passed twice (once as
part of the script, and once from the parse out). So this PR fixes that
problem by not passing it as part of the e2e script commands.

For more details, see https://github.com/knative/serving/issues/12788

Fixes https://github.com/knative/serving/issues/12788

/assign @dprotaso

### DIFF
--- a/e2e-tests.sh
+++ b/e2e-tests.sh
@@ -152,7 +152,12 @@ function initialize() {
         # Skip parsed flag (and possibly argument) and continue
         # Also save it to it's passed through to the test script
         for ((i=1;i<=skip;i++)); do
-          e2e_script_command+=("$1")
+          case "$1" in
+            # Don't need to pass enable-ha to the test script
+            # For details, see https://github.com/knative/serving/issues/12788
+            --enable-ha) : ;;
+            *) e2e_script_command+=("$1") ;;
+          esac
           shift
         done
         continue
--- a/e2e-tests.sh
+++ b/e2e-tests.sh
@@ -137,13 +137,17 @@ CLOUD_PROVIDER="gke"
 function initialize() {
   local run_tests=0
   local custom_flags=()
+  local parse_script_flags=0
   E2E_SCRIPT="$(get_canonical_path "$0")"
   local e2e_script_command=( "${E2E_SCRIPT}" "--run-tests" )
+
+  for i in "$@"; do
+    if [[ $i == "--run-tests" ]]; then parse_script_flags=1; fi
+  done
 
   cd "${REPO_ROOT_DIR}"
   while [[ $# -ne 0 ]]; do
     local parameter=$1
-    # TODO(chizhg): remove parse_flags logic if no repos are using it.
     # Try parsing flag as a custom one.
     if function_exists parse_flags; then
       parse_flags "$@"
@@ -152,12 +156,10 @@ function initialize() {
         # Skip parsed flag (and possibly argument) and continue
         # Also save it to it's passed through to the test script
         for ((i=1;i<=skip;i++)); do
-          case "$1" in
-            # Don't need to pass enable-ha to the test script
-            # For details, see https://github.com/knative/serving/issues/12788
-            --enable-ha) : ;;
-            *) e2e_script_command+=("$1") ;;
-          esac
+          # Avoid double-parsing 
+          if (( parse_script_flags )); then
+            e2e_script_command+=("$1")
+          fi
           shift
         done
         continue


e2e-tests.sh
```diff
@@ -152,7 +152,12 @@ function initialize() {
         # Skip parsed flag (and possibly argument) and continue
         # Also save it to it's passed through to the test script
         for ((i=1;i<=skip;i++)); do
-          e2e_script_command+=("$1")
+          case "$1" in
+            # Don't need to pass enable-ha to the test script
+            # For details, see https://github.com/knative/serving/issues/12788
+            --enable-ha) : ;;
```
dprotaso, line 158: This seems too specific to serving.  ie. this doesn't solve the issue if we pass the `--perf` flag the same way as `--enable-ha`

This whole script is a bit strange -  initialize will create the cluster (kntest/kubetest) and invoke itself again which is hard to reason about.

I wonder if what we really want is to do an initial pass for the flags that this script cares about and avoid calling `parse_flags` in the first pass. Then ensure we`parse_flags` in the second pass.

psschwei: Thinking out loud, but what do you think about adding a `--parse-flags` (or `--parse-script-flags`) for the second pass and then only run `parse_flags` if that flag is present?

dprotaso: Technically the presence of `--run-tests` could be signal whether we call l `parse-flags`

But it's unclear to me what happens if we don't remove these extra flags when creating the cluster - ie. will it fail saying unknown flag/argument etc.

psschwei: I'm checking in https://github.com/knative/serving/pull/12896 ... so far so good

psschwei: Well, it works in that the tests pass.... but we end up losing other flags that get passed in on the original command. Which is _probably_ not good...

For example, anything passed to the initial presubmit-tests.sh call: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_serving/12896/istio-latest-no-mesh_serving_main/1521497009092038656#1:build-log.txt%3A5

dprotaso: > Which is probably not good...

cc @chizhg @kvmware for input

psschwei: Well... maybe it's not a problem.... the `parse_flags` function sets environmental variables in the `e2e-common.sh` script. I _think_ they're being set on the first call (should be able to check that in the PR above...)

kvmware: I am not familiar with the execution flow here, but from a quick pass it seems calling itself is a bad idea:
1. problem we having now
2. recursion hurts my head
Could we have a lighter weight e2e entry script that simply passes it's flags (removing any particular to it) to the actual test? Essentially the `--run-tests` logic should be decoupled, I think. Ignore me if this all sounds stupid, like I said new to this bit here.

psschwei: > maybe it's not a problem

I was right, the envvars get set in the first pass, so not needed again: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_serving/12896/istio-latest-no-mesh_serving_main/1521514608752332800#1:build-log.txt%3A89 

(I guess if they WERE being set twice, we would've got similar read-only errors as we were getting with enable-ha...)

psschwei: >     2. recursion hurts my head

+1

> Could we have a lighter weight e2e entry script that simply passes it's flags (removing any particular to it) to the actual test? Essentially the `--run-tests` logic should be decoupled, I think. Ignore me if this all sounds stupid, like I said new to this bit here.

I'm not super familiar with the flow either, so while in general I'd say yes, I'm not sure how much effort that would be... looking (very quickly) at the CI, it doesn't look like any of the parsed flags are used on the second call, so maybe that piece can be dropped?

e2e-tests.sh
```diff
@@ -137,13 +137,17 @@ CLOUD_PROVIDER="gke"
 function initialize() {
   local run_tests=0
   local custom_flags=()
+  local parse_script_flags=0
   E2E_SCRIPT="$(get_canonical_path "$0")"
   local e2e_script_command=( "${E2E_SCRIPT}" "--run-tests" )
 
+  for i in "$@"; do
+    if [[ $i == "--run-tests" ]]; then parse_script_flags=1; fi
+  done
```
dprotaso, line 146: nit: can drop the for loop

```suggestion
  [[ "$@" == *"--run-tests"* ]] && parse_script_flags=1
```

chizhg: I think the for loop is better. Your approach will also match something like `--run-tests-setup` which could cause errors that are hard to debug.

chizhg:
LGTM from me.
/lgtm

/hold
In case Dave and Krsna have other comments.

e2e-tests.sh
dprotaso: good point

dprotaso:
/lgtm
/hold cancel

re: Krsna's comment about this whole flow - I think it's out of scope for this PR and can be accomplished in a follow up re-arch of these scripts

PR #179, opened 2022-05-23
Set Go version by update-deps.sh

nak3:
Currently Go version is not consistent and every repository needs to bump the go version in `go.mod` manually like these commits:

https://github.com/knative-sandbox/net-kourier/commit/53f2bec99c447c4d5d332a89f894664befb3c128
https://github.com/knative-sandbox/net-certmanager/commit/8f622b7cbfb1829edb4863ce89930f0cf9673a1c

This patch sets the go version in go.mod automatically by `update-deps.sh`.

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -61,6 +61,8 @@ if [[ -z "${ARTIFACTS:-}" ]]; then
   export ARTIFACTS="$(mktemp -d)"
 fi
 
+export GOVERSION="${GOVERSION:-1.17}"
+
 # On a Prow job, redirect stderr to stdout so it's synchronously added to log
 (( IS_PROW )) && exec 2>&1
 
@@ -631,6 +633,9 @@ function go_update_deps() {
       echo "Nothing to upgrade."
     fi
   fi
+
+  group "Set Go version"
+  go mod edit -go=${GOVERSION}
 
   group "Go mod tidy and vendor"
 


nak3:
/hold

This is not urgent PR so should be merged after the 1.5 release.

library.sh
```diff
@@ -632,6 +634,9 @@ function go_update_deps() {
     fi
   fi
 
+  group "Set Go version"
+  go mod edit -go=${GOVERSION}
```
skonto, line 638: I guess other repos dont have issues with moving to 1.17. So better do it for all.

dprotaso:
Also the `go` version in the mod file is just a min supported go version - and doesn't necessarily need to be bumped.

1.17 is an exception because the Go Team decided to include all the transient dependences in a second require block for module pruning.

nak3:
> Also the go version in the mod file is just a min supported go version - and doesn't necessarily need to be bumped.
>
> 1.17 is an exception because the Go Team decided to include all the transient dependences in a second require block for module pruning.

Yeah, I also think that the `go` version in the mod file is not so important and unnecessary to be cared. That's why we can modify it by auto script. If Go team makes some exception in the future like 1.17, we again have to modify the mod file one by one :sweat:

nak3:
Anyway, I don't have strong opinion to automate this one. Closing.

PR #189, opened 2022-06-28
Add [skip-dot-release] feature

kvmware:
# Changes
- :gift: Adds the ability for one to add `[skip-dot-release]` to a commit message to skip that commit when calculating if a dot release needs to be built

Fixes: https://github.com/knative/test-infra/issues/3386


release.sh
```diff
@@ -233,8 +233,8 @@ function prepare_dot_release() {
   setup_branch
   # Use the original tag (ie. potentially with a knative- prefix) when determining the last version commit sha
   local github_tag="$(hub_tool release | grep "${last_version}")"
-  local last_release_commit="$(git rev-list -n 1 "${github_tag}")"
-  local release_branch_commit="$(git rev-list -n 1 upstream/"${RELEASE_BRANCH}")"
+  local last_release_commit="$(git rev-list --invert-grep --grep "\[skip-dot-release\]" -n 1 "${github_tag}")"
```
dprotaso, line 236: I don't think we need the grep when looking up the `last_release_commit`

dprotaso: Can we update the log output on ~240 to include all three SHAs  or maybe print a message the release branch SHA has skipped commits to just surface the behaviour

I can forsee someone not knowing about this functionality and wondering why a release isn't happening (I hope there is a low chance of that)

kvmware: `I don't think we need the grep when looking up the last_release_commit`
If we don't then we would be potentially matching a patched dot_release with a skip with another skip dot release aka they wouldnt match and a dot release would be done

`...I can forsee someone not knowing about this functionality...`
yea, I was worried about that too. Will do :+1:

kvmware: Done. Added some more info on the message as well.

PR #196, opened 2022-07-14
fix typos and upstream library.sh changes

upodroid:
Part of https://github.com/knative/infra/issues/134

I'm putting this in a separate branch so the changes can be synced by knobots and avoid shipping this to other repos via go module.

Will merge this to main once all repos have been fixed.

/cc @kvmware @evankanderson

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -632,31 +632,13 @@ function go_update_deps() {
     fi
   fi
 
-  group "Go mod tidy and vendor"
+  group "Go mod tidy"
 
   # Prune modules.
   local orig_pipefail_opt=$(shopt -p -o pipefail)
   set -o pipefail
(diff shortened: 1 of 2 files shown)


library.sh
```diff
@@ -632,31 +632,13 @@ function go_update_deps() {
...
-    -o -name "BUILD.bazel" \
-    -o -name "*_test.go" \) -exec rm -f {} +
-
-  export GOFLAGS=-mod=vendor
-
-  group "Updating licenses"
-  update_licenses third_party/VENDOR-LICENSE "./..."
-
-  group "Removing broken symlinks"
-  remove_broken_symlinks ./vendor
```
cardil, line 659: I think this is still needed, as not all repos are vendorless.

PR #198, opened 2022-07-18
Start Signing our Releases

upodroid:
This PR enables signing our releases.

Initially, I will enroll nighly jobs for net-certmanager(images) and kn-plugin-admin(binaries) to test signing images and binaries. Once we are happy, we can look at a wider rollout.

I need to modify the nightly prowjobs to add the flag `--signing-key=projects/knative-nightly/locations/global/keyRings/cosign/cryptoKeys/cosign` to the release.sh script

/cc @kvmware @cardil @evankanderson


evankanderson:
I presume you mean "the combination of those repos covers both binaries and images+manifests patterns" -- each one is one pattern; I'm not sure we have a repo which does both styles of release.

release.sh
```diff
@@ -301,6 +302,31 @@ function build_from_source() {
   if [[ $? -ne 0 ]]; then
     abort "error building the release"
   fi
+  sign_release
+  # Do not use `||` above or any error will be swallowed.
```
evankanderson, line 306: Why use this pattern and not:

```
sign_release || abort "error signing the release"
```

?

release.sh
```diff
@@ -301,6 +302,31 @@ function build_from_source() {
...
+  fi
+}
+
+# Build a release from source.
+function sign_release() {
+  ## Sign the images with cosign
+  ## For now, check if ko has created imagerefs.txt file. In the future, missing image refs will break
+  ## the release for all jobs that publish images.
+  if [[ -f "imagerefs.txt"  && -z ${SIGNING_KEY} ]]; then
+      echo "Signing Images with the key ${SIGNING_KEY}"
```
evankanderson, line 318: It looks like this means we'll be signing the images but not the manifests that reference the images. Should we be signing those as well?

upodroid:
That is right. Updating PR body.

release.sh
```diff
@@ -301,6 +302,31 @@ function build_from_source() {
...
+      COSIGN_EXPERIMENTAL=1 cosign sign $(cat image.refs) --key gcpkms://"${SIGNING_KEY}"
```
evankanderson, line 319: Where does this output go / get stored?

I see the next output being redirected to a file and saved.

evankanderson: Is `image.refs` supposed to be the same string as `imagerefs.txt`?

evankanderson (commented):
Who has access to the signing keys? How are they published / curated?

release.sh
upodroid: Running `cosign sign` signs the image and pushes it to the registry.



> Is `image.refs` supposed to be the same string as `imagerefs.txt`?

Yup, typo.

release.sh
upodroid: Copied previous approach.

upodroid:
> Who has access to the signing keys? How are they published / curated?

The private key is stored on GCP KMS and access is restricted to specific principals.

https://github.com/knative/test-infra/blob/main/infra/gcp/nightly/cosign.tf

All operations will be logged by Google.

release.sh
upodroid: I don't think it is needed. Endusers are most like going to do something like this: https://kyverno.io/docs/writing-policies/verify-images/ so only the image should be signed for now.

release.sh
upodroid: Fixed

release.sh
puerco: I agree but in the kubernetes images, we chose to sign everything that gets tagged. We sign the index as there are situations (say in a multiarch cluster) where you may choose to do the verification on the index instead of a single image. Another reason is that at some point, you may want to attach sboms to the images so it makes sense to have the index signed to validate the index sbom too.

(Hint `cosign sign --recursive` is your friend here)

release.sh
```diff
@@ -301,6 +302,27 @@ function build_from_source() {
...
+}
+
+# Build a release from source.
+function sign_release() {
+  ## Sign the images with cosign
+  ## For now, check if ko has created imagerefs.txt file. In the future, missing image refs will break
+  ## the release for all jobs that publish images.
+  if [[ -f "imagerefs.txt"  && -z ${SIGNING_KEY} ]]; then
+      echo "Signing Images with the key ${SIGNING_KEY}"
+      COSIGN_EXPERIMENTAL=1 cosign sign $(cat imagerefs.txt) --key gcpkms://"${SIGNING_KEY}"
```
puerco, line 315: OK, let me do an argument here.

Of course, signing with any KMS works but I would strongly suggest moving towards keyless signing now instead of switching later. The main reason here is that you are effectively buying the project a key management problem it does not really need.

Each release will now require to have its own keys published with the artifacts (as you are doing it now by copying it to the bucket), and while this would not be that bad 10 years ago but now we have a lot of automated tooling everywhere verifying signatures which expect to find the certificate attached to the image. 

You can use the key file or the cloud kms to verify the images, but most consumers will go via the automated route and it will simply not work for them. The kyverno policy at the top of the doc you linked above is a good example: You could potentially need one such policy listing each release if you rotate the keys with regularity (which you should). Compare that with the [keyless policy in the same doc](https://kyverno.io/docs/writing-policies/verify-images/#keyless-signing-and-verification), it is a single one to verify all releases.

So tldr, keyless should be straightforward to implement and saves the project and its users some problems now but a ton in the future.

puerco (commented):
Based on the experience and discussions we've had implementing signing for the kubernetes org, I have two comments on the current approach :point_down:

release.sh
upodroid: Thanks for the comments, i'll go ahead with Keyless signing.

Infra is being created at https://github.com/knative/test-infra/pull/3516

kvmware:
LGTM

puerco:
This is great @upodroid, in the long run it will be much easier for consumers of the images. Since we're on the topic of signing what are you thinking about signing the rest of the release artifacts?

upodroid:
A few repos ship binaries such as https://github.com/knative/client and https://github.com/knative-sandbox/kn-plugin-func/releases. For those, I'll be signing the checksums that are generated, the logic is in the PR. Something like https://goreleaser.com/install/#binaries

I still have the CNCF wide problem around code signing and apple notary that was talked about at this week's SIG Release meeting.

upodroid:
This is ready. Complete run is available at https://github.com/knative-sandbox/net-certmanager/pull/440

upodroid:
/hold cancel

thanks

PR #204, opened 2022-07-25
drop warning in `go_run`

dprotaso:
the warning text goes to stdout which breaks subshell invocations

/assign @cardil


cardil (approved):
/lgtm

Okay. I'll see if we could restore the warn in non breaking way, ie. pushing it to stderr instead.

PR #205, opened 2022-07-27
:gift: Support multiple Golang modules in a single project

### LINKED ISSUE #85: update-deps script should support repos with multiple go.mod files
The `update-deps` script do not properly support repos that have multiple go modules (`go.mod` files).

You can see an example of such project in https://github.com/knative-sandbox/kn-plugin-event/pull/76. Only the root `go.mod` file is processed. I think the script should look up all `go.mod` files (outside of vendor dir) and process them all.

cardil:
# Changes

- :gift: Support multiple Golang modules in a single project

/kind enhancement

Fixes #85


upodroid:
+1

We need to have a dummy github repo that we execute a our complete release process on to test these changes. 

Changes in this repo broke the 1.7 release which didn't flag up in the nightlies release process.

upodroid:
You also don't want to be patching hack repo on release day as every repo needs to be bumped then resynced properly.

cardil:
> We need to have a dummy github repo that we execute a our complete release process on to test these changes.

There is a `test/` directory that serves as such a dummy project. There are some tests, and I'm adding some with this PR as well. But, that's not enough, as you said @upodroid. We can add more tests there... (Writing them in Go, instead of bash could also help)

library.sh
```diff
@@ -126,7 +126,7 @@ function subheader() {
 
 # Simple warning banner for logging purposes.
 function warning() {
-  make_banner "!" "$1"
```
kvmware, line 129: are we using the `make_banner` function somewhere else if not that can be removed

library.sh
```diff
@@ -679,14 +706,16 @@ function go_update_deps() {
   update_licenses third_party/VENDOR-LICENSE "./..."
 
   group "Removing broken symlinks"
-  remove_broken_symlinks ./vendor
+  remove_broken_symlinks vendor
```
kvmware, line 709: is there a reason for removing the `./` prefix here? usually in shell scripts it is conventional to leave that to indicate this is a local directory

library.sh
cardil: The make_banner writes to a stdout. I wanted the warning to write to stderr, so when executed with capturing the output the warning will not influence the command.

The following will still work, but the deprecated warning will be captured in the build log.

```shell
deps="$(run_go_tool knative.dev/test-infra/buoy@latest buoy float ./go.mod "${buoyArgs[@]}")"
```

kvmware: Shouldn't you keep the original `make_banner "!" "$1"` but just pipe it to stderr then?

library.sh
cardil: No big reason, just a simplification.

The `./` prefix isn't really required. It doesn't save us from anything, but might be cumbersome if tried to concatenate paths, for example.

kvmware: Unless we are doing the concat i would keep the `./` just for convention reasons, but just my 2cents not a blocker

library.sh
cardil: Hmm, I'll test that. I don't know it that'll work...

cardil:
> _@cardil: We can add more tests there... (Writing them in Go, instead of bash could also help)_

The follow-up PR #215 is doing just that...

mgencur:
This PR looks good to me. I was first confused why there are changes in schema/third_party and schema/vendor but @cardil told me it was generated by running update-deps on this same repo. Personally, I would include it at least in a separate commit with  the title "Run update-deps.sh for this repo", or maybe even include this in a follow up PR. But having this here is ok as well.

kvmware:
/lgtm
/approve
/hold
Unhold when you ready to observe the changes @cardil

cardil:
/unhold

Let's merge. It should be fine. Will react if otherwise...

PR #215, opened 2022-09-02
:broom: Rewrite unit tests in Golang

### LINKED ISSUE #61: Scripts should survive `set -Eeuo pipefail` mode
Bash scripts in knative.dev/hack should survive using strict mode of:

```
set -Eeuo pipefail
```

Description:
* `-E` - _If set, any trap on ERR is inherited by shell functions, command substitutions, and commands executed in a subshell environment._
* `-e` - _Exit immediately if a pipeline (see Pipelines), which may consist of a single simple command (see Simple Commands), a list (see Lists), or a compound command (see Compound Commands) returns a non-zero status._
* `-e` - _Treat unset variables and parameters other than the special parameters ‘@’ or ‘*’ as an error when performing parameter expansion._
* `-o pipefail` - _If set, the return value of a pipeline is the value of the last (rightmost) command to exit with a non-zero status, or zero if all commands in the pipeline exit successfully._

Ref.: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

Failure to that forces users to write bad code.

Here's an example of what happens if such strict mode is used before sourcing the hack library: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-sandbox_eventing-kafka/467/pull-knative-sandbox-eventing-kafka-integration-test-channel-consolidated/1377237757516058624

cardil:
# Changes

- :broom: Rewrite unit tests in Golang
- :broom: Fix scripts to survive in the `set -Eeuo pipefail` mode

/kind cleanup

Fixes #61


cardil:
https://github.com/kubernetes/test-infra/issues/27358 blocks this PR

cardil:
https://github.com/kubernetes/test-infra/issues/27358 blocks this PR

/hold

cardil:
/unhold

https://github.com/kubernetes/test-infra/issues/27358 wasn't the cause of the unit test failure.

It was due to: 

https://github.com/knative/hack/blob/f9dc7220d36264f5aa2a304b3663612815b757bc/library.sh#L70

It was fixed in 8ba8bc0e2d71d17f4d77cf949ea25f5712a05404

cardil:
I'd like this merged, before I could work on https://github.com/knative/hack/issues/221

upodroid (approved):
Thanks

PR #223, opened 2022-09-08
:gift: Use modscope tool to get info about Go modules

### LINKED ISSUE #220: The `go.work` should be respected
When a project has a `go.work` file, and that file lists modules, only those modules should be considered by the scripts.

If the project doesn't contain a `go.work` file, we should only act on the project root directory.

Follows up on #85

cardil:
# Changes

- :gift: Use modscope tool to get info about Go modules

/kind enhancement

Fixes #220


kvmware:
LGTM, but I would like to you to run the shell scripts being changed through shellcheck https://www.shellcheck.net/ or use the action `reusable-style.yaml` from actions repo

cardil:
I run shellcheck in my IDE, already.

upodroid (approved):
LGTM

PR #224, opened 2022-09-08
Add flag to output image references

upodroid:
Instead of going to 30+ repos to add this flag, I can add it here. `ko` will write the image references to a file for `cosign` to sign.

/kind enhancement
/cc @kvmware


kvmware:
smart
/lgtm

PR #230, opened 2022-09-26
Deprecate calling kubetest2 via kntest

upodroid:
# Changes

- deprecate calling kubetest2 via kntest. It will take a while for the change to propagate.
- test PR with the change: https://github.com/knative/serving/pull/13329

Fixes #

/cc @cardil

### DIFF
--- a/e2e-tests.sh
+++ b/e2e-tests.sh
@@ -176,10 +176,6 @@ function initialize() {
     shift
   done
 
-  if [[ "${CLOUD_PROVIDER}" == "gke" ]]; then
-      extra_gcloud_flags+=("--addons=NodeLocalDNS")
-  fi
-
   readonly SKIP_DUMP_ON_FAILURE
   readonly TEARDOWN
   readonly CLOUD_PROVIDER
--- a/e2e-tests.sh
+++ b/e2e-tests.sh
@@ -133,6 +133,7 @@ E2E_SCRIPT=""
 function initialize() {
   local run_tests=0
   local custom_flags=()
+  local extra_gcloud_flags=()
   local parse_script_flags=0
   E2E_SCRIPT="$(get_canonical_path "$0")"
   local e2e_script_command=( "${E2E_SCRIPT}" "--run-tests" )
@@ -178,7 +179,7 @@ function initialize() {
   done
 
   if [[ "${CLOUD_PROVIDER}" == "gke" ]]; then
-      custom_flags+=("--addons=NodeLocalDNS")
+      extra_gcloud_flags+=("--addons=NodeLocalDNS")
   fi
 
   readonly SKIP_DUMP_ON_FAILURE
@@ -186,7 +187,7 @@ function initialize() {
   readonly CLOUD_PROVIDER
 
   if (( ! run_tests )); then
-    create_test_cluster "${CLOUD_PROVIDER}" custom_flags e2e_script_command
+    create_test_cluster "${CLOUD_PROVIDER}" custom_flags extra_gcloud_flags e2e_script_command
   else
     setup_test_cluster
   fi
--- a/e2e-tests.sh
+++ b/e2e-tests.sh
@@ -19,8 +19,6 @@
 
 source "$(dirname "${BASH_SOURCE[0]:-$0}")/infra-library.sh"
 
-readonly TEST_RESULT_FILE=/tmp/${REPO_NAME}-e2e-result
-
 # Tear down the test resources.
 function teardown_test_resources() {
   header "Tearing down test environment"
--- a/infra-library.sh
+++ b/infra-library.sh
@@ -131,6 +131,10 @@ function create_gke_test_cluster() {
     _extra_gcloud_flags+=("--logging=NONE --monitoring=NONE")
   fi
 
+  if [[ "${CLOUD_PROVIDER}" == "gke" ]]; then
+      extra_gcloud_flags+=("--addons=NodeLocalDNS")
+  fi
+
   if [[ "${ENABLE_PREEMPTIBLE_NODES:-}" == "true" ]]; then
     _extra_gcloud_flags+=("--preemptible")
   fi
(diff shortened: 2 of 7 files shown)


infra-library.sh
```diff
@@ -117,23 +117,37 @@ function create_kind_test_cluster() {
...
   fi
 
   if [[ "${ENABLE_PREEMPTIBLE_NODES:-}" == "true" ]]; then
-    extra_gcloud_flags="${extra_gcloud_flags} --preemptible"
+    _extra_gcloud_flags="${_extra_gcloud_flags} --preemptible"
+  fi
+  
+  if [[ "${USE_KUBETEST2_DIRECTLY:-}" == "true" ]]; then
+    _extra_gcloud_flags="${_extra_gcloud_flags[@]} --quiet"
+    kubetest2 gke "${_custom_flags[@]}" --up --down \
```
cardil, line 140: I would like to avoid people needing to install an appropriate version of kubetest2 on their machines.

I'm okay to use kubetest2 directly on PROW, but on developer machines it should be called via:

```shell
go_run sigs.k8s.io/kubetest2@v0.0.0-20220903000027-a46ae8d7fa37 gke "${_custom_flags[@]}" --up --down \
```

cardil:
Unit tests needs to be updated. They are failing now because they were expecting a call to kntest.

infra-library.sh
upodroid: kubetest2 can't be run by developers as it needs access to GCP and it will try to talk to boskos endpoint which is isn't reachable from dev laptops.

upodroid: https://github.com/knative/test-infra/issues/2681

cardil: Of course, it could. People could use different GCP project, right?

dprotaso:
/hold 

I see some test is commented out

e2e-tests.sh
```diff
@@ -178,15 +177,15 @@ function initialize() {
   done
 
   if [[ "${CLOUD_PROVIDER}" == "gke" ]]; then
-      custom_flags+=("--addons=NodeLocalDNS")
+      extra_gcloud_flags+=("--addons=NodeLocalDNS")
   fi
```
dprotaso, line 181: Maybe move this to `create_gke_test_cluster()` in `infra-library.sh` and drop `extra_gcloud_flags` env var here

infra-library.sh
```diff
@@ -117,29 +117,38 @@ function create_kind_test_cluster() {
...
-    --extra-gcloud-flags="${extra_gcloud_flags}"
+  if [[ ! " ${_custom_flags[*]} " =~ "--machine-type=" ]]; then
+      _custom_flags+=("--machine-type=e2-standard-4")
+  fi
+  kubetest2 gke "${_custom_flags[@]}" --rundir-in-artifacts --up --down \
+    --boskos-heartbeat-interval-seconds=20 --v=1 \
+    --network=e2e-network --boskos-acquire-timeout-seconds=1200 \
+    --region="${E2E_CLUSTER_REGION},us-east1,us-west1" --gcloud-extra-flags="${_extra_gcloud_flags[*]}" \
+    --retryable-error-patterns='.*does not have enough resources available to fulfill.*,.*only \\d+ nodes out of \\d+ have registered; this is likely due to Nodes failing to start correctly.*,.*All cluster resources were brought up.+ but: component .+ from endpoint .+ is unhealthy.*' \
+    --test=exec -- "${_test_command[@]}"
```
dprotaso, line 153: nit: can we put an arg on each line so it's easier to read

performance-tests.sh
```diff
@@ -76,7 +76,8 @@ EOF
     update_knative || abort "failed to update knative"
   fi
   # get benchmark name from the cluster name
-  local benchmark_name=$(get_benchmark_name "$1")
+  local benchmark_name
+  benchmark_name=$(get_benchmark_name "$1")
```
dprotaso, line 80: lol - i hate this lint rule haha

test/unit/presubmit_test.go
```diff
@@ -43,23 +42,25 @@ func TestMainFunc(t *testing.T) {
...
+	// 			"--enable-workload-identity --cluster-version=latest "+
+	// 			"--extra-gcloud-flags --enable-stackdriver-kubernetes "+
+	// 			"--no-enable-ip-alias --no-enable-autoupgrade "+
+	// 			"--addons=NodeLocalDNS "+
+	// 			"--test-command=%s/test/e2e-tests.sh "+
+	// 			"--run-tests --extra-gcloud-flags= --logging=NONE "+
+	// 			"--monitoring=NONE", rootDir)),
+	// 		contains("INTEGRATION TESTS PASSED"),
+	// 	},
+	// }}
```
dprotaso, line 62: Should this be deleted or changed?

performance-tests.sh
upodroid: yeah

test/unit/presubmit_test.go
upodroid: Hmm, so kntest prints the kubetest2 args to stdout and allowed this test to ensure the flags were tested here. The upstream kubetest2 doesn't do this so the test is broken. I might be able to fix that upstream and reenable this test.

dprotaso: I'd say just delete this assertion then

upodroid: Cool

upodroid: done

e2e-tests.sh
```diff
@@ -133,6 +131,7 @@ E2E_SCRIPT=""
 function initialize() {
   local run_tests=0
   local custom_flags=()
+  local extra_gcloud_flags=()
```
dprotaso, line 134: nit: remove this var

dprotaso (approved):
/lgtm
/approve
/hold 

holding for nit - drop if you don't want to do it in this PR

e2e-tests.sh
upodroid: I'll revisit this in a separate PR

PR #238, opened 2022-10-17
:gift: Use Knative ls-tags tool

cardil:
# Changes

- :gift: Use Knative ls-tags tool

/kind enhancement

Related https://github.com/knative/actions/pull/72

**Release Note**

```release-note
Use Knative ls-tags tool
```

### DIFF
--- a/test/unit/sharedlib_test.go
+++ b/test/unit/sharedlib_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"embed"
 	"fmt"
+	"io"
 	"math/rand"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


test/unit/sharedlib_test.go
```diff
@@ -341,6 +360,63 @@ func (s shellScript) write(t TestingT, src string) string {
...
+
+type fnPrefetcher func(t TestingT)
+
+func (f fnPrefetcher) prefetch(t TestingT) {
+	f(t)
+}
+
+func goRunHelpPrefetcher(tool string) prefetcher {
+	return fnPrefetcher(func(t TestingT) {
+		c := exec.Command("go", "run", tool, "--help")
```
jrangelramos, line 375: why you're calling help here! Could you please clarify better the purpose of the pre-fetcher

cardil: I added that prefetcher code to the tests, so to make sure the tools are already downloaded and compiled. Without it, the tests might fetch those tools during runtime, and that was causing tests to have unexpected output: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_hack/238/unit-tests_hack_main/1582291191687286784

Maybe I'll add some comments to the source code to make it clear.

cardil:
/test unit-tests

We should add prefetching of test-infra, as when it updates the test might fail.

cardil:
IDK why such error happens:
```
error: could not lock config file .git/config: File exists
```
/test unit-tests

upodroid:
Seen that before, it is a flaky behaviour of the build system

PR #243, opened 2022-11-02
drop support for the istio add on flag

dprotaso:
This thing is so old and we don't use it so drop it.

In the past we would be install the Istio add-on available in GKE but we ended up installing it ourselves.


dprotaso:
I don't understand the error in the unit test cc @cardil

cardil:
The unit tests expects the script output to have:

```
kubectl wait job --for=condition=Complete --all -n istio-system --timeout=5m
```

I guess you should adjust that assertion appropriately.

dprotaso:
> The unit tests expect the script output to have:

Ah - the error message in the test could have been more clear

upodroid (approved):
LGTM

PR #249, opened 2022-11-10
Trap calls are now executed in LIFO order

dprotaso:
We tend to add traps at various points in our test scripts.
The assumption we have is the trap will be executed when
external resources are in a certain state.

ie.
1. Setup Cluster - trap delete_cluster
2. Setup Knative - trap teardown_knative
3. Setup Tests   - trap dump_test_failure

Prior traps are executed in FIFO order so the cluster
would be torn down before the test dump would occur.

In LIFO order the test dump would occur first, then the
teardown of knative and then finally the cluster.

/assign @cardil 
/assign @upodroid

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -647,7 +647,7 @@ function add_trap {
     local current_trap
     current_trap="$(trap -p "$trap_signal" | cut -d\' -f2)"
     local new_cmd="($cmd)"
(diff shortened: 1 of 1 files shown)


cardil (commented):
Would you mind adding a code comment with your justification from pr description?

dprotaso:
Actually I think the trap command should be LIFO and that would solve this problem

dprotaso:
Ok - done please take a look

PR #261, opened 2023-01-03
Upload attestations and print cosign version

upodroid:
This change uploads the attesttation payload so we can view it with the job it was built from and we can see more clearly the cosign version we used.

/cc @kvmware


upodroid:
this can be merged now

PR #274, opened 2023-02-16
Add shell linter github action

### LINKED ISSUE #269: Add shellcheck linter
This repo is mostly bash scripts and shellcheck helps us have secure good bash.

Our actions repo has a shellcheck linter reusable github action that should be used.

kauana:
# Changes
- Add shellcheck linter as a github action (uses knative's reusable [style action](https://github.com/knative/actions/blob/main/.github/workflows/reusable-style.yaml))
- See [this PR ](https://github.com/kauana/hack/pull/2) for an example of how the action works when someone attempts to add bad shell code

/kind enhancement

Fixes #269

### DIFF
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -4,9 +4,13 @@ on:
   push:
   pull_request:
     branches:
-    - 'main'
-    - 'release-*'
+      - "main"
+      - "release-*"
 
 jobs:
   style:
     uses: knative/actions/.github/workflows/reusable-style.yaml@main
+    with:
+      yaml: false
+      gofmt: false
+      gha_yaml: false
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -1,0 +1,7 @@
+name: Linter
+
+on: [pull_request]
+
+jobs:
+  style:
+    uses: knative/actions/.github/workflows/reusable-style.yaml@main
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -1,6 +1,11 @@
 name: Linter
 
-on: [pull_request]
+on:
+  push:
+  pull_request:
+    branches:
+    - 'main'
+    - 'release-*'
 
 jobs:
   style:


.github/workflows/linter.yaml
```diff
@@ -0,0 +1,7 @@
+name: Linter
+
+on: [pull_request]
```
cardil, line 3: Please change it to:

```suggestion
on:
  push:
    branches:
    - 'main'
    - 'release-*'
  pull_request:
    branches:
    - 'main'
    - 'release-*'
```

but first, please change it to the following, so we can test those changes:

```suggestion
on:
  push:
  pull_request:
    branches:
    - 'main'
    - 'release-*'
```

This will allow this job to execute, on your commit already. So, we will see the outcome.

Unfortunately, I fear it's not that simple. There are multiple shellcheck violations in this repository. Simply adding a check is a no-go, as it will block any development until all issues are resolved. The reusable-style check does look only on changed files, but still that may be too much to handle.

That's why I ask you to simulate simple change in the scripts (add/change a comment to some existing function, for example [here](https://github.com/knative/hack/blob/d71d569c430876fa2d0f4d3e355ace20975fe376/release.sh#L124)). If we get errors from the whole file, that's a problem. Issues only in vicinity of the change (+/- 10 lines for example are ok).

Some linters also support the to-do files. Those are files holding the exclusions of issues. The idea is to adopt a linter on code that isn't clean, and gradually clean in, removing more and more of the TODO file (Described here https://scottm.medium.com/rubocop-in-legacy-projects-part-1-todos-and-todonts-877ace9f23b7). Shellcheck doesn't support that AFAIK.

Our shellcheck linter seams to utilize the reviewdog, which should handle that for us. Unfortunately, it's set in `file` mode, instead of the default `diff_context`. See: https://github.com/knative/actions/blob/6d02b89684cc3f5f8e06e655e26f8a5c0cfa126b/composite/style/shell/action.yml#L29

cardil (changes requested):
Thanks for your PR! :heart_eyes: 

Please take a look at my comment, about the scope of the linter.

Adding a hold until we are sure it's safe to adopt it

/hold

.github/workflows/linter.yaml
kauana: Thank you for the thorough feedback @cardil! Unfortunately, it doesn't seem that the action is running on this PR even after adding your suggestion 🤔 

In the meantime, I will experiment changing a file like you mentioned on my fork instead.

kauana: Okay, I've done some experimentation on my other [PR](https://github.com/kauana/hack/pull/2). 

Some learnings:
- using `on.push` makes the changes job fail ([see here](https://github.com/kauana/hack/actions/runs/4197770487/jobs/7280561519)). Issue goes away when on push gets removed ([see here](https://github.com/kauana/hack/actions/runs/4197798990/jobs/7280596067)). This is probably related to some workflow permission.

You were correct that the action was failing for the wrong reason. After fixing it, the linter never fails, but it does print warnings for modified script files. The action is configured as follows:
```
    INPUT_GITHUB_TOKEN: ***
    INPUT_TOOL_NAME: shellcheck
    INPUT_LEVEL: warning
    INPUT_FILTER_MODE: file
    INPUT_FAIL_ON_ERROR: false
    INPUT_REVIEWDOG_FLAGS: 
    INPUT_CLEANUP: true
```

I believe giving warnings and not failing is a good behavior for now as it will encourage folks to fix their script without blocking any PRs to this repo, but I would like to hear your thoughts @cardil, thanks!

kauana: Hey @cardil, any thoughts on the above? Thanks

kvmware:
Hey there, sorry for this hanging here for so long. I think the correct solution is to add hack repo to [knobots](https://github.com/knative-sandbox/knobots) to use the style directive. That way the workflow is managed by the automation system. There is examples of the style reusable action in serving, eventing, etc. that uses this system.

kauana:
Hey @kvmware, sorry for the wait as well! I believe `knative/hack` is already added to knobots, but the github actions are currently [excluded for this repo](https://github.com/knative-sandbox/knobots/blob/main/actions-exclude.yaml#LL1). 

If that's ok with you, I'm happy to update that so `knative/hack` can take advance of the style action. Thanks

kvmware:
That should work thanks

Issue #272, opened 2023-02-13
The release script fails to publish proper checksums and their signatures


cardil:
As evidenced by [kn-event 1.9.1 release job](https://prow.knative.dev/view/gs/knative-prow/logs/release_kn-plugin-event_release-1.9_periodic/1625125552232861696#1:build-log.txt%3A1492-1494), the release script fails to publish a proper checksum file, together with (`.sig` and `.pem`) files.

The files at gs bucket are valid, but the `checksums.txt` file published at Github releases was invalid.

I suspect the stale `checksums.txt` (before MacOS signing) was published on Github releases, and that caused the upload command to fail.

/kind bug

cardil:
I've downloaded all the artifacts from the Github to compare them with gs bucket. It looks like all binaries are not matching the ones from the bucket:

```shell
$ diff -u latest github
Binary files latest/kn-event-darwin-amd64 and github/kn-event-darwin-amd64 differ
Binary files latest/kn-event-darwin-arm64 and github/kn-event-darwin-arm64 differ
Binary files latest/kn-event-linux-amd64 and github/kn-event-linux-amd64 differ
Binary files latest/kn-event-linux-arm64 and github/kn-event-linux-arm64 differ
Binary files latest/kn-event-linux-ppc64le and github/kn-event-linux-ppc64le differ
Binary files latest/kn-event-linux-s390x and github/kn-event-linux-s390x differ
Binary files latest/kn-event-windows-amd64 and github/kn-event-windows-amd64 differ
```

cardil:
The changes done in https://github.com/knative/hack/pull/268 are not enough to make an errorless release. Further work needs to be done.

cardil:
For whatever reason, the script doesn't find the `checksums.txt` file produced by kn-event build:

https://prow.knative.dev/view/gs/knative-prow/logs/release_kn-plugin-event_release-1.9_periodic/1625125552232861696#1:build-log.txt%3A1301

As a result, the `$ARTIFACTS_TO_PUBLISH` variable contains two `checksums.txt` files, and that breaks the artifact to upload to the GitHub page (GH API responds the name of the artifact is duplicated).

PR #275, opened 2023-02-16
:bug: Find checksums file works with ARTIFACTS_TO_PUBLISH variable

cardil:
# Changes

- :bug: Find checksums file works with ARTIFACTS_TO_PUBLISH variable

/kind bug

Fixes https://github.com/knative/hack/issues/272
Follow up to https://github.com/knative/hack/pull/268

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -328,12 +328,18 @@ function get_images_in_yamls() {
   fi
 }
 
+# Finds a checksums file within the given list of artifacts (space delimited)
+# Parameters: $n - artifact files
 function find_checksums_file() {
-  for file in "$@"; do
-    if [[ "${file}" == *"checksums.txt" ]]; then
-      echo "${file}"
-      return 0
-    fi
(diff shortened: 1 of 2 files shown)


upodroid (approved):
Not a blocker, but what is the kn-event release script doing that is different from the other kn-* scripts?

/hold

cardil:
> @upodroid: Not a blocker, but what is the kn-event release script doing that is different from the other kn-* scripts?

The kn-plugin-repo is unique in a couple of places. I often use it as a proving grounds for new development approaches.

A couple of differences from the top of my head:

- it uses mostly the Go-native built tools (via [magetasks](https://github.com/wavesoftware/go-magetasks)), instead of bash scripts
- have multiple `go.mod` files and a `go.work`
- builds artifacts and a checksums file in `build` subdirectory

evankanderson (commented):
/lgtm
/approve

PR #284, opened 2023-04-17
Fix bug that broke signatures

upodroid:
/cc @cardil @kvmware @BobyMCbobs 

The change in #268 broke image signatures. I'm also copying the image refs and attestations to the artifacts folder for easy verification.

I need to write a tool a to easily check sigstore signatures on all the nightly jobs.


cardil (approved):
/lgtm

Those pesky $@

cardil:
Thanks for the fix @upodroid!

PR #291, opened 2023-05-23
Install Istio for e2e tests

### LINKED ISSUE #290: kind-tests_hack_main job failing
The [e2e kind test](https://github.com/knative/hack/blob/main/test/e2e-kind.sh) has been failing because it times out waiting for `net-istio` to be installed ([example](https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_hack/289/kind-tests_hack_main/1658875324210876416)).

kauana:
# Changes

- Installs Istio (ever since the istio addon stopped being supported, Istio hasn't been installed). The [recent probe additions to net-istio](https://github.com/knative-sandbox/net-istio/pull/1095) probably caused the tests to fail because Istio wasn't installed.
- Follows the same flow used by [knative/client](https://github.com/knative/client/blob/main/test/common.sh#L32)

Fixes #290 
Related issue #269

### DIFF
--- a/test/e2e-kind.sh
+++ b/test/e2e-kind.sh
@@ -18,7 +18,49 @@ set -Eeo pipefail
 
 source "$(dirname "${BASH_SOURCE[0]:-$0}")/../e2e-tests.sh"
 
+export INGRESS_CLASS=${INGRESS_CLASS:-istio.ingress.networking.knative.dev}
+
+function is_ingress_class() {
+  [[ "${INGRESS_CLASS}" == *"${1}"* ]]
+}
+
+# Copied from https://github.com/knative/client/blob/main/test/common.sh#L32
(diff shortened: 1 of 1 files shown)


library.sh
```diff
@@ -536,7 +536,7 @@ function start_knative_serving() {
   echo "Installing Serving core components from $2"
   kubectl apply -f "$2"
   echo "Installing net-istio components from $3"
-  kubectl apply -f "$3"
+  kubectl apply -f  -l knative.dev/crd-install=true "$3"
```
dprotaso, line 539: `start_knative_serving` assumes you have istio installed so I think we shouldn't add this label selector

dprotaso: This will skip installing the net-istio controller etc

PR #297, opened 2023-07-07
Replace test-infra with toolbox

nak3:
As per title, test-infra repo https://github.com/knative/test-infra has been closed so use toolsbox instead.


nak3:
Fixed except for kntest due to https://github.com/knative/toolbox/issues/14

nak3:
The code style is not related this PR. It should be fixed in a separated PR if we fix them.

upodroid (approved):
LGTM

Issue #310, opened 2023-09-01
Hack scripts shouldn't enforce vendor dir


cardil:
Knative hack scripts need to work with repos with vendor and without it.

The scripts code should be adjusted to allow both situations.

/kind enhancement

PR #311, opened 2023-09-05
:gift: Don't force the vendoring of deps

cardil:
# Changes

- :gift: Don't force the vendoring of deps

/kind enhancement

Fixes https://github.com/knative/hack/issues/310
Requires https://github.com/knative/hack/pull/222

**Release Note**

```release-note
The vendor directory isn't enforced in Knative projects anymore
```


go.work
```diff
@@ -4,4 +4,5 @@ use (
 	.
 	schema
 	test
+	test/vendorproj
```
cardil, line 7: For the record: I added this submodule to test if the scripts allow having a mixed project, with some modules having `vendor/` and some don't.

PR #316, opened 2023-09-21
:bug: Collect LICENSE files from all dependencies, even without vendor dir

### LINKED ISSUE #315: update_licenses should still run when repos don't have vendor folder
see discussion here: https://github.com/knative/pkg/pull/2810#discussion_r1319110291

Caused by #311

cardil:
# Changes

- :bug: Collect LICENSE files from all dependencies, even without vendor dir

/kind bug

Fixes #315

**Release Note**

```release-note
The license files are collected, even if a project doesn't have the vendor directory
```


go.work.sum
dprotaso, line 1: was this committed by accident?

cardil: No. Those are sums on the workspace level. It's the same concept as sums on module level.

However, I'm not entirely sure how they are used...

cardil: Here's some info: https://stackoverflow.com/a/71613130

It's a bit of a useless thing, at least my understanding, but still it's a part of go tooling...

upodroid (approved):

go.work.sum
cardil: What's more confusing for me, is that this file was updated after I executed our update deps script, which runs `go mod tidy` in every module. This should sync proper deps in modules and the workspace file should be empty.

Maybe the update deps script should delete the `go.work.sum` file before the loop?!?

dprotaso: I'm sorta mixed on whether we should check in go.work files - there seems to be a discussion here

https://github.com/golang/go/issues/53502

dprotaso: hack repo since it has with multiple modules seems like a candidate :shrug:

PR #317, opened 2023-09-21
:bug: The `go.work.sum` don't influence the dependencies resolution

cardil:
If it's not needed. If the file isn't empty after the update, it should be committed.

Some info: https://stackoverflow.com/a/71613130

# Changes

- :bug: The `go.work.sum` don't influence the dependencies resolution

/kind bug

Follow up #316


cardil:
Some info: https://stackoverflow.com/a/71613130

cardil:
What's strange, when I open the project in Goland IDE, or invoke a "Sync Go Module" operation in that IDE, the `go.work.sum` is being recreated with the contents of #316.

The Goland IDE must call some `go` command, but I don't know which one.

dprotaso:
There really shouldn't be any dependencies in the hack repo - the original intention was for it to just propagate shell scripts to other repos

Now the deps used here will have an influence of the version selection (might not be a big issue) - but I would encourage moving tools to toolbox and test fixtures can be generated programmatically or put under `testdata` (eg. test vendor project)

dprotaso:
Also note this dep is now in the stdlib 

https://github.com/knative/hack/blob/c4a34c34512e6aee3d8bfcc0e311fb8cd7aa2cf4/go.mod#L6 


https://pkg.go.dev/errors

cardil:
> @dprotaso There really shouldn't be any dependencies in the hack repo - the original intention was for it to just propagate shell scripts to other repos
> 
> Now the deps used here will have an influence of the version selection (might not be a big issue) - but I would encourage moving tools to toolbox and test fixtures can be generated programmatically or put under `testdata` (eg. test vendor project)

Good point. I think this is an oversight on my side, while doing the https://github.com/knative/hack/pull/222. Before that PR there ware go deps, but only in isolated go sub modules.

We could move the `cmd/script` to the toolbox, or just create an additional `go.mod` in cmd.

Let's capture this in a separate issue.

cardil:
@dprotaso the issue https://github.com/knative/hack/issues/318

upodroid (approved):

PR #320, opened 2023-09-22
:bug: Remove deps from go.mod for knative.dev/hack

### LINKED ISSUE #318: The `knative.dev/hack` go package shouldn't contain any go deps
The knative.dev/hack go package shouldn't contain any go deps because it is being sourced in all repos in `hack/tools.go` file, like seen below. Having deps in this package could influence the deps in other consuming repos unintentionally.

```go
import (
  _ "knative.dev/hack"
)
```

Caused by #222
More info https://github.com/knative/hack/pull/317#issuecomment-1730266629

/kind bug

cardil:
# Changes

-  :bug: Remove deps from go.mod for knative.dev/hack

/kind bug

Fixes #318


upodroid (approved):
Thanks

PR #338, opened 2023-10-25
Make generate-groups.sh and generate-knative.sh executable conditionally

mgencur:
The codegen-library.sh script might be executed within a container that doesn't run as root in which case changing permissions is not allowed and the whole scripts fails.
This change gives opportunity to set the permissions in advance (e.g. during docker build) and then run this script successfully.

# Changes

- Make files executable only if they're not executable yet

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


pierDipi (approved):
/lgtm
/approve

I've double-checked it with:
```
$ [ -x /bin/sh ]
$ echo $?
```

PR #339, opened 2023-11-02
add function to set latest semver to latest

kvmware:
# Changes
- :gift: Add function to release.sh to set highest semver as latest in github

Fixes https://github.com/knative/infra/issues/201

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -668,8 +668,10 @@ function set_latest_to_highest_semver() {
   fi
   local last_version="$(echo "${releases}" | grep '^v[0-9]\+\.[0-9]\+\.[0-9]\+$' | sort -r -V | head -1)"
   local release_id # don't combine with the line below, or $? will be 0
-  [[ $? -eq 0 ]] || abort "cannot get relase id from github"
   release_id="$(hub_tool api /repos/${ORG_NAME}/${REPO_NAME}/releases/tags/knative-${last_version} | jq .id)"
+  if [[ $? -ne 0 ]]; then
+    abort "cannot get relase id from github"
+  fi
   hub_tool api --method PATCH /repos/knative/serving/releases/$release_id -F make_latest=true > /dev/null || abort "error settomg $last_version to 'latest'"
   echo "Github release ${last_version} set as 'latest'"
 }
--- a/release.sh
+++ b/release.sh
@@ -655,6 +655,23 @@ function publish_artifacts() {
   banner "New release published successfully"
 }
 
+# Sets the github release with the highest semver to 'latest'
+function set_latest_to_highest_semver() {
+  # List latest release
+  local releases # don't combine with the line below, or $? will be 0
+  # Support tags in two formats
+  # - knative-v1.0.0
+  # - v1.0.0
+  releases="$(hub_tool -p release | cut -d '-' -f2)"
+  [[ $? -eq 0 ]] || abort "cannot list releases"
+  local last_version="$(echo "${releases}" | grep '^v[0-9]\+\.[0-9]\+\.[0-9]\+$' | sort -r -V | head -1)"
+  local release_id # don't combine with the line below, or $? will be 0
+  [[ $? -eq 0 ]] || abort "cannot get relase id from github"
+  release_id="$(hub_tool api /repos/${ORG_NAME}/${REPO_NAME}/releases/tags/knative-${last_version} | jq .id)"
+  hub_tool api --method PATCH /repos/knative/serving/releases/$release_id -F make_latest=true > /dev/null || abort "error settomg $last_version to 'latest'"
+  echo "Github release ${last_version} set as 'latest'"
+}
+
 # Entry point for a release script.
 function main() {
   parse_flags "$@"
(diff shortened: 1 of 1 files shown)


kvmware:
This assumes that `jq` is available.

dprotaso:
/approve

will let others add lgtm

kvmware:
https://github.com/knative/infra/blob/main/images/prow-tests/Dockerfile#L65

release.sh
```diff
@@ -655,6 +655,23 @@ function publish_artifacts() {
...
 }
 
+# Sets the github release with the highest semver to 'latest'
+function set_latest_to_highest_semver() {
+  # List latest release
+  local releases # don't combine with the line below, or $? will be 0
+  # Support tags in two formats
+  # - knative-v1.0.0
+  # - v1.0.0
+  releases="$(hub_tool -p release | cut -d '-' -f2)"
```
cardil, line 665: This will fail if `set -e` is set. It should be set.

Also, storing array in a regular variable isn't a good idea.

release.sh
```diff
@@ -655,6 +655,23 @@ function publish_artifacts() {
...
+  [[ $? -eq 0 ]] || abort "cannot list releases"
```
cardil, line 666:
```suggestion
  if ! [[ $? -eq 0 ]]; then
    abort "cannot list releases"
  fi
```

release.sh
```diff
@@ -655,6 +655,23 @@ function publish_artifacts() {
...
+  local last_version="$(echo "${releases}" | grep '^v[0-9]\+\.[0-9]\+\.[0-9]\+$' | sort -r -V | head -1)"
+  local release_id # don't combine with the line below, or $? will be 0
+  [[ $? -eq 0 ]] || abort "cannot get relase id from github"
+  release_id="$(hub_tool api /repos/${ORG_NAME}/${REPO_NAME}/releases/tags/knative-${last_version} | jq .id)"
```
cardil, line 670: This seems to be the other way around

```suggestion
  release_id="$(hub_tool api /repos/${ORG_NAME}/${REPO_NAME}/releases/tags/knative-${last_version} | jq .id)"
  if [[ $? -ne 0 ]]; then
    abort "cannot get relase id from github"
  fi
```

cardil (changes requested):

release.sh
kvmware: This (minus the `-p`) is copy/pasted from https://github.com/knative/hack/blob/main/release.sh#L220 which has been working....

cardil: I'm not referring to `-p`.

You can't check for `[[ $? -eq 0 ]]` unless `set +e` is active. Otherwise, every failed command exit the script (as it should).

cardil: I think this could be simplified to:

```bash
local last_version
last_version="$(hub_tool -p release | cut -d'-' -f2 | grep '^v[0-9]\+\.[0-9]\+\.[0-9]\+$'| sort -r -V | head -1)"
```

release.sh
```diff
@@ -655,6 +655,25 @@ function publish_artifacts() {
...
+  # List latest release
+  local releases # don't combine with the line below, or $? will be 0
+  # Support tags in two formats
+  # - knative-v1.0.0
+  # - v1.0.0
+  releases="$(hub_tool -p release | cut -d '-' -f2)"
+  if ! [[ $? -eq 0 ]]; then
+    abort "cannot list releases"
+  fi
+  local last_version="$(echo "${releases}" | grep '^v[0-9]\+\.[0-9]\+\.[0-9]\+$' | sort -r -V | head -1)"
```
cardil, line 669:
```bash
local xxx="$(false)"
```
is always true. https://www.shellcheck.net/wiki/SC2155

release.sh
kvmware: yes, but not as readable imo. technically we can "simplify" lots of this and make it a one liner monster :P

release.sh
kvmware: which is fine for this line. that is why other variables don't have it all one one line and have comments

release.sh
cardil: Bash isn't readable. Writing invalid Bash, just to be readable, is not a good idea.

See this article: https://stackoverflow.com/a/17602863/844449

kvmware: how is this invalid bash?

kvmware: oh, because you are saying array in a var is bad practice... I wouldn't call it invalid but i see your point

cardil: invalid bash == one that has bugs in it

release.sh
cardil: How is this fine? If any command within that expression, we don't get fail but an empty value as last_version. We expect if any command fails, the script fails.

cardil (approved):
/lgtm

Those conditional failures could be dropped, as the `set -e` should enforce the script will fail at first error, but I can accept this as a defensive touch. They will ensure this works as intended even if this is executed with set +e (unfortunately it might be)

Thanks @kvmware for fixing this!

Issue #342, opened 2023-11-07
`set_latest_to_highest_semver` func fails the nighly release jobs


cardil:
https://github.com/knative/hack/blob/0d99b7ee9d63bcb9e8611707c919497bb304dc86/release.sh#L748C31-L748C31

By calling to `set_latest_to_highest_semver` func fails the nightly jobs:

https://prow.knative.dev/view/gs/knative-prow/logs/nightly_func_main_periodic/1721817559931555840

/kind bug

Caused by https://github.com/knative/hack/pull/339

cardil:
This isn't fixed, still...

/reopen

cardil:
See example failure: https://prow.knative.dev/view/gs/knative-prow/logs/nightly_eventing-kafka-broker_main_periodic/1722547880884965376

PR #346, opened 2023-11-09
:bug: Set latest release only when publishing to Github

cardil:
# Changes

- :bug: Set latest release only when publishing to Github

/kind bug

Follow up https://github.com/knative/hack/pull/343
Fixes https://github.com/knative/hack/issues/342

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -658,13 +658,18 @@ function publish_artifacts() {
 
 # Sets the github release with the highest semver to 'latest'
 function set_latest_to_highest_semver() {
-  local last_version # don't combine with the line below, or $? will be 0
+  if ! (( PUBLISH_TO_GITHUB )); then
+    return 0
+  fi
+  echo "Setting latest release to highest semver"
+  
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -658,13 +658,18 @@ function publish_artifacts() {
 
 # Sets the github release with the highest semver to 'latest'
 function set_latest_to_highest_semver() {
-  local last_version # don't combine with the line below, or $? will be 0
+  if ! (( PUBLISH_TO_GITHUB )); then
```
kvmware, line 661: if we move this call to the publish to github function then that function already has this logic there

release.sh
```diff
@@ -658,13 +658,18 @@ function publish_artifacts() {
...
+    return 0
+  fi
+  echo "Setting latest release to highest semver"
+  
+  local last_version release_id
```
kvmware, line 666:
```suggestion
  local last_version release_id  # don't combine with assignment else $? will be 0
```

PR #357, opened 2024-01-08
include additional k8s bash file when updating exec permission

dprotaso:
There have been changes to the codegen scripts in k8s - we now need to add an exec bit to the following file

```
k8s.io/code-generator/generate-internal-groups.sh
```


upodroid (approved):
LGTM

PR #369, opened 2024-02-21
drop toolchain directive

dprotaso:
Toolchain directive isn't necessary - and it's breaking go-licenses.

This is a workaround that drops it when updating dependencies.


dprotaso:
tested this with serving seems good - `go mod vendor` doesn't add it back

/hold cancel

PR #373, opened 2024-03-13
switch default region to avoid instability

dprotaso:
GKE folks recommend using a different region

Fixes: https://github.com/knative/serving/issues/14935


upodroid (approved):
Interesting suggestion.

PR #375, opened 2024-03-27
:lady_beetle: Don't set latest release when not publishing to GH

cardil:
# Changes

- :lady_beetle: Don't set latest release when not publishing to GH

/kind bug

Fixes TBD


cardil:
Sorry, for the empty PR. This is already in.

Closing...

PR #376, opened 2024-03-27
:broom: Check licenses only, but not bundle them

cardil:
# Changes

- :broom: Check licenses only, but not bundle them

/kind cleanup

Fixes knative/community#1441

**Release Note**

```release-note
License files are no longer bundled with Knative sources, but only checked for validity
```

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -770,8 +770,8 @@ function __go_update_deps_for_module() {
     remove_broken_symlinks ./vendor
   fi
 
-  group "Updating licenses"
-  update_licenses third_party/VENDOR-LICENSE "./..."
+  group "Checking licenses"
+  check_licenses
   )
 }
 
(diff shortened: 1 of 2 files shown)


library.sh
```diff
@@ -770,8 +770,8 @@ function __go_update_deps_for_module() {
     remove_broken_symlinks ./vendor
   fi
 
-  group "Updating licenses"
-  update_licenses third_party/VENDOR-LICENSE "./..."
+  group "Checking licenses"
+  check_licenses
```
cardil, line 774: Maybe we should actively delete those `LICENSE` files, instead of relying on of every project contributors to do it manually?

dprotaso: yeah sounds good

dprotaso: actually - that might break builds

dprotaso: because we need to stop the symlink to third party licenses

cardil: What might break builds? The removal of the LICENSE files?

cardil: I like the idea @dprotaso proposed, to add the deletion of those licenses in the next release.

PR #389, opened 2024-08-01
Remove geo replication for GCR

upodroid:
We have been doing geo-replication for a long time, but it's costly and we don't tell our end-users about it.

It costs about 4x to store images in Artifact Registry vs Container Registry. 0.10 USD vs 0.026 USD for a multiregion GCS bucket. 

image

Storing in just gcr.io/knative-nightly which is the US multi-region should be sufficient.

/cc @dprotaso @cardil


dprotaso:
Note - for the changes to take effect in release branches downstream repos need to update their `knative.dev/hack` dependency

dprotaso:
Also @upodroid to save on cost should we delete the images in those regions?

upodroid:
I will delete the registries in knative-nightly project today.

I haven't migrated the prod one yet.

PR #397, opened 2024-09-04
Fix premature codegen cleanup exit on 1

dsimansk:
# Changes

- Fix premature codegen cleanup exit on 1
- Fix `truncate` param to be macos compatible

/cc @skonto @cardil

### DIFF
--- a/codegen-library.sh
+++ b/codegen-library.sh
@@ -133,17 +133,20 @@ function restore-changes-if-its-copyright-year-only() {
   local difflist
   log "Cleaning up generated code"
   difflist="$(mktemp)"
-  git diff --exit-code --name-only > "$difflist"
-  # list git changes and skip those which differ only in the boilerplate year
-  while read -r file; do
-    # check if the file contains just the change in the boilerplate year
-    if [ "$(LANG=C git diff --exit-code --shortstat -- "$file")" = ' 1 file changed, 1 insertion(+), 1 deletion(-)' ] && \
-        [[ "$(git diff --exit-code -U1 -- "$file" | grep -Ec '^[+-]\s*[*#]?\s*Copyright 2[0-9]{3}')" -eq 2 ]]; then
-      # restore changes to that file
-      git checkout -- "$file"
-    fi
-  done < "$difflist"
-  rm -f "$difflist"
+  if ! git diff --exit-code --name-only > /dev/null; then
+    # list git changes and skip those which differ only in the boilerplate year
+    git diff --name-only > "$difflist"  
+    while read -r file; do
+      # check if the file contains just the change in the boilerplate year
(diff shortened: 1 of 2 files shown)


dsimansk:
This should fix failed dep updates like: https://github.com/knative-extensions/knobots/actions/runs/10683503406/job/29611769106

codegen-library.sh
```diff
@@ -133,16 +133,18 @@ function restore-changes-if-its-copyright-year-only() {
   local difflist
   log "Cleaning up generated code"
   difflist="$(mktemp)"
-  git diff --exit-code --name-only > "$difflist"
```
skonto, line 136: I guess this was killing the script?

dsimansk: Yep, it was exiting with `1` not continuing to actual content check. Perhaps we could narrow down the conditions, i.e. only check boilerplate copyright in `zz_generated*.go` files. I'm not sure if it behaves differently in various Bash instances, but wrapping it into the if should be safer for now.

cardil: In shell, when you set `exit at error` flag (which you should), every command that returns non 0 will break the script.

Unless it is used in conditional statement, like: `if` or combined with `||` or `&&`.

codegen-library.sh
```diff
@@ -133,16 +133,18 @@ function restore-changes-if-its-copyright-year-only() {
...
+    # list git changes and skip those which differ only in the boilerplate year
+    git diff --name-only > "$difflist"  
+    while read -r file; do
+      # check if the file contains just the change in the boilerplate year
+      if [ "$(LANG=C git diff --exit-code --shortstat -- "$file")" = ' 1 file changed, 1 insertion(+), 1 deletion(-)' ] && \
+          [[ "$(git diff --exit-code -U1 -- "$file" | grep -Ec '^[+-]\s*[*#]?\s*Copyright 2[0-9]{3}')" -eq 2 ]]; then
+        # restore changes to that file
+        git checkout -- "$file"
+      fi
+    done < "$difflist"
```
cardil, line 146:
```suggestion
    while read -r file; do
      # check if the file contains just the change in the boilerplate year
      if [ "$(LANG=C git diff --exit-code --shortstat -- "$file")" = ' 1 file changed, 1 insertion(+), 1 deletion(-)' ] && \
          [[ "$(git diff --exit-code -U1 -- "$file" | grep -Ec '^[+-]\s*[*#]?\s*Copyright 2[0-9]{3}')" -eq 2 ]]; then
        # restore changes to that file
        git checkout -- "$file"
      fi
    done < <(git diff --name-only)
```

codegen-library.sh
cardil: Yes, this is a bug. And the fix looks good to me. Thanks!

cardil (approved):
/lgtm

PR #401, opened 2024-10-23
Export KO_FLAGS for consuming scripts

dsimansk:
# Changes

- Export `KO_FLAGS` for consuming scripts

Per the Slack thread with Vincent.  If you prefer to do it globally. Otherwise we could also isolate the changes as needed into consuming `${knative-repo}/hack/release.sh`.

/cc @dprotaso @upodroid @cardil


cardil (approved):
/lgtm

I have no problems with exporting this

PR #413, opened 2025-02-17
Bump default GKE version to 1.31

creydr:
As per title, needed for 1.18 (https://github.com/knative/pkg/blob/ae2c6bcc1f5968766febc343903819b43d7ec903/version/version.go#L36)


creydr:
@cardil @upodroid could you PTAL?

upodroid (approved):

PR #470, opened 2026-04-16
Add env var to pass extra flag to license check, parametrize tool ver…

dsimansk:
…sions

# Changes

- Add env var to pass extra flag to license check
- Parametrize tool versions

After `go-licenses` v2 the following error is present in repos using `gotest.tools` library. The issue is license format, that was previously recognized in v1 (v1 check passes) as header type, and I think it should be the case for v2. The `go-licenses` is currently unmaintained, so we can either stick with v1 version or keep using v2 with an extra flags `--ignore` when needed.  

I'll open an issue with them.

v1: https://github.com/google/go-licenses/blob/v1.6.0/licenses/classifier.go#L95-L98
v2: https://github.com/google/go-licenses/blob/v2.0.1/licenses/classifier.go#L69-L71
Adding the following fixes the issue:
```
if match.MatchType != "License" &&  match.MatchType != "Header" {
```

https://github.com/gotestyourself/gotest.tools/blob/main/LICENSE
```
  Did not find license for library 'gotest.tools/v3/assert'.
  Did not find license for library 'gotest.tools/v3/assert/cmp'.
  Did not find license for library 'gotest.tools/v3/internal/assert'.
  Did not find license for library 'gotest.tools/v3/internal/format'.
  Did not find license for library 'gotest.tools/v3/internal/source'.

```

Consuming repo example: 
```
# Library gotest.tools/v3 which has a short-form Apache 2.0 notice
# that go-licenses v2 fails to classify.
export GO_LICENSES_FLAGS="--ignore gotest.tools/v3"
```

Client PR: https://github.com/knative/client/pull/2191

/cc @dprotaso

### DIFF
--- a/library.sh
+++ b/library.sh
@@ -35,6 +35,13 @@ if [[ ! -v GOPATH ]]; then
   fi
 fi
 
+# Pinned tool versions
+readonly GUM_VERSION="v0.14.1"
+readonly GOTESTSUM_VERSION="v1.13.0"
+readonly GOTESTFMT_VERSION="v2.5.0"
+readonly TERMINAL_TO_HTML_VERSION="v3.10.0"
+readonly GO_LICENSES_VERSION="v2.0.1"
+
 # Useful environment variables
 [[ -v PROW_JOB_ID ]] && IS_PROW=1 || IS_PROW=0
 readonly IS_PROW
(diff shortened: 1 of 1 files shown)


dsimansk:
Tracking issue: https://github.com/google/go-licenses/issues/339

dsimansk:
@dprotaso we should probably cherry it release branches too, wdyt?

dprotaso:
@dsimansk sure go for it

PR #475, opened 2026-09-17
Add go workspace support to bump go toolchain

dsimansk:
# Changes

- Add go workspace support to bump go toolchain

```release-note
Add go workspace support to bump go toolchain
```


dsimansk:
This change should fix: https://github.com/knative/pkg/actions/runs/35202556079/job/105140531345?pr=3392


```
    go: module . listed in go.work file requires go >= 1.26.0, but go.work lists go 1.25.0; to update it:
    	go work use
```

dsimansk:
It won't work for vendorless repositoritories, because go.mod go.work mismatch prevents also:

```
source "$(go run knative.dev/hack/cmd/script library.sh)"
```

/close

